PR #1, opened 2025-03-16
initial files

corrideat:


src/internals.ts
```diff
@@ -0,0 +1,2344 @@
...
+              hash: v.keyShareHash
+            })
+          })
+        }
+      },
+      [SPMessage.OP_PROP_DEL]: notImplemented,
+      [SPMessage.OP_PROP_SET] (v: SPOpPropSet) {
+        if (!state._vm.props) config.reactiveSet(state._vm, 'props', Object.create(null))
+        config.reactiveSet(state._vm.props!, v.key, v.value)
+      },
```
taoeffect, line 1107: Apparently this used to not use `reactiveSet`: https://github.com/okTurtles/libcheloniajs/commit/df33723564474dc91832955ab38e5985283359f3#diff-cc2f6524fbb0853be076a033b87a685aa551d36b4a8e201f97eebed4921c6dd9L1103-L1105

Ricardo could you please make a list of all non-type source related changes you've made and please port them over to Group Income first?

This PR shouldn't modify any source. Before we switch over to this library in Group Income, there may still be changes we need to make to Chelonia because of bugs like https://github.com/okTurtles/group-income/issues/2797 - so please avoid making any source-code changes in this PR that are not type related (make them in a PR to GI instead).

src/utils.ts
```diff
@@ -0,0 +1,878 @@
...
+    if (key.meta?.keyRequest?.contractID && findSuitableSecretKeyId(state, [SPMessage.OP_KEY_ADD], ['sig'])) {
+      const data = unwrapMaybeEncryptedData(key.meta.keyRequest.contractID)
+
+      // Are we subscribed to this contract?
+      // If we are not subscribed to the contract, we don't set pendingKeyRequests because we don't need that contract's state
+      // Setting pendingKeyRequests in these cases could result in issues
+      // when a corresponding OP_KEY_SHARE is received, which could trigger subscribing to this previously unsubscribed to contract
+      if (data && internalSideEffectStack) {
+        const keyRequestContractID = data.data
+        const reference = unwrapMaybeEncryptedData(key.meta.keyRequest.reference)
```
taoeffect, line 357: Another source related change

src/internals.ts
corrideat: This is one of the few non-type changes that I've made. However, this one has no impact as we're not using this opcode.

corrideat: Actually, I think that this may need to be reverted, since `reactiveSet` is used afterwards when the state is applied.

corrideat: > This PR shouldn't modify any source. 

I agree in principle, but, this change being an exception, some source changes are in fact type-related changes, which fix bugs that weren't previously caught.

taoeffect:
I think, before merging this, we should pass every single file through an LLM comparing it to the original file and ask the LLM if there are any differences between the files unrelated to type signature changes.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,21 @@
+name: Node.js Test
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
```
taoeffect, line 7: This repo uses `main` because Github decided to change the default for some reason.

I can either change it to `master` (but then the PR would need to be resubmitted), or this needs to be updated. Feel free to make the call.

corrideat: Indeed.

corrideat:
Regarding the rest of the feedback:

The changes to `eventsBetween` are related to fetching chatroom messages in both directions. This is part of the branch I'm currently working on in https://github.com/okTurtles/group-income/pull/2841. This change needs to be ported to PR https://github.com/okTurtles/group-income/pull/2825 once this is merged.

The `use strict` directive isn't needed because it's automatically added by TypeScript where appropriate (note the files under `dist/cjs` have it; it's not needed for ESM because ESM is exclusively strict mode).

The `handleFetchResult` function was imported from GI. Since this is a separate library, `handleFetchResult` needed to be copied over (however, we could avoid duplicating code by importing it from GI).

.github/workflows/ci.yml
corrideat: BTW, note that you can adjust the name of the default branch on your GitHub preferences.

src/chelonia.ts
```diff
@@ -0,0 +1,2068 @@
...
+    return data
+  },
+  'chelonia/out/latestHEADInfo': function (this: CheloniaContext, contractID: string) {
+    return this.config.fetch(`${this.config.connectionURL}/latestHEADinfo/${contractID}`, {
+      cache: 'no-store',
+      signal: this.abortController.signal
+    }).then(handleFetchResult('json'))
+  },
+  'chelonia/out/eventsAfter': eventsAfter,
+  'chelonia/out/eventsBefore': function (this: CheloniaContext, contractID: string, beforeHeight: number, limit: number, options: { stream: boolean }) {
```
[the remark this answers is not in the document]
corrideat: This would be a more involved change, as those functions are used in other places besides `ChatMain.vue` (in particular, `eventsAfter` is used in Chelonia itself). Should I go ahead with making the change to named parameters, or should I partially revert the changes to `eventsBetween` (and if the latter, please provide the desired signature)?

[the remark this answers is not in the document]
taoeffect: The APIs should be consistent... so I don't know what you mean by "revert". The problem with `eventsBetween` before was a real problem. One solution to it would be to move those optional params into an `options` object like `eventsBefore` has. Another solution (that you picked) would be to move all of them (except `contractID`) into an object.

Reverting to how it was before isn't an option because it was really poorly designed before... So if by "revert" you mean "go with the solution where optional params are placed in an `options` object", that's fine too. But if by "revert" you mean, "undo our improvement"... I don't think that's a good idea.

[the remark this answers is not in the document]
corrideat: I'm agreeing with that they should ideally be consistent. OTOH, it looks like I was overzealous with the way I implemented named parameters.

So, there are two ways I see to address the inconsistency issue:

1. Make the same change to `eventsAfter` and `eventsBefore`, i.e., making everything after `contractID` a named parameter.
2. _Partially_ revert the change to `eventsBetween`. This would in practice mean keeping some arguments positional (as they were) and some named (other than `stream`)

Both alternatives are pretty simple, but the first option (making everything named) will require more care to ensure no mistakes are introduced (e.g., by forgetting to update a call to the new signature).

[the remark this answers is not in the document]
corrideat: Another question (I guess it applies to either option, since both are breaking changes):

Would you rather I update the code here (in this repo) and in https://github.com/okTurtles/group-income/pull/2825, or that I make a separate PR to https://github.com/okTurtles/group-income with the same API changes (this way, https://github.com/okTurtles/group-income/pull/2825 will continue to have just `import`-related changes)?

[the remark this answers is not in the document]
corrideat: https://github.com/okTurtles/group-income/pull/2850

PR #11, opened 2025-09-15
Add disableAutoDedup

corrideat:
Add the ability to disable deduplication checks when recreating messages. The current behaviour prevents sending unnecessary or redundant opcodes, but it could be undesirable in some cases, especially when using `OP_ATOMIC`, since the de-duplication logic isn't smart enough to detect some valid operations that should be sent as specified (for example, `OP_ATOMIC[OP_KEY_DEL k1, OP_KEY_ADD k1]` will result in just `OP_ATOMIC[OP_KEY_DEL k1]` as the logic thinks that `k1` already exists, but neglects to see that it was deleted by the first `OP_KEY_DEL`).


corrideat:
See https://github.com/okTurtles/libcheloniajs/pull/10, which has a very similar check (that PR was to disable it). It's probably redundant to check for the same thing in two locations, and likely we don't need that logic in `keyAdd`, so long as recreateEvent is always run (which it should be, except for `OP_CONTRACT`)

PR #16, opened 2025-10-18
Whitespace

corrideat:


src/db.ts
```diff
@@ -131,22 +138,34 @@ export default sbp('sbp/selectors/register', {
...
+      )
     }
   },
   'chelonia/db/addEntry': function (entry: SPMessage): Promise<string> {
     // because addEntry contains multiple awaits - we want to make sure it gets executed
     // "atomically" to minimize the chance of a contract fork
     return sbp('okTurtles.eventQueue/queueEvent', `chelonia/db/${entry.contractID()}`, [
-      'chelonia/private/db/addEntry', entry
+      'chelonia/private/db/addEntry',
+      entry
```
taoeffect: I'd say the entire PR looks good to me, with this being the only eyebrow-raising change. Personally think it looked better before and see no reason to split it across lines like this.

src/encryptedData.test.ts
```diff
@@ -58,15 +62,21 @@ describe('Encrypted data API', () => {
...
+            }
           }
         }
-      }
-    } as ChelContractState, serializedEncryptedData, 0, { [id]: key })
+      } as ChelContractState,
+      serializedEncryptedData,
+      0,
+      { [id]: key }
+    )
```
taoeffect: Well, I suppose these types of changes aren't really needed either.

taoeffect: No big deal either way I suppose. But note that prettier doesn't need to be run again.. And really we don't need to add it as a project dependency

taoeffect: Indeed you're right 👍

PR #18, opened 2025-10-19
deserializedHEAD API

### LINKED ISSUE #17: Expose a way to get a deserializedHEAD given a hash
## Problem

`eventsBetween` implements fetching a hash and to retrieve its height. We sometimes need to fetch this information directly.

## Solution

Implement a dedicated API for this.

corrideat:
Closes #17

### DIFF
--- a/src/chelonia.ts
+++ b/src/chelonia.ts
@@ -1461,12 +1461,16 @@ export default sbp('sbp/selectors/register', {
       })
       .then(handleFetchResult('json'))
   },
-  'chelonia/out/deserializedHEAD': async function (this: CheloniaContext, hash: string, { contractID }: { contractID?: string } = {}) {
+  'chelonia/out/deserializedHEAD': async function (
+    this: CheloniaContext, hash: string, { contractID }: { contractID?: string } = {}
+  ) {
+    // contractID is optional because this selector could be used for looking up
+    // a contractID given a hash.
     const message = await sbp('chelonia/out/fetchResource', hash, {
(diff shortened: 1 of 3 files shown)


src/chelonia.ts
```diff
@@ -1498,16 +1508,7 @@ export default sbp('sbp/selectors/register', {
...
-          .fetch(`${this.config.connectionURL}/file/${startHash}`, {
-            signal: this.abortController.signal
-          })
-          .then(handleFetchResult('text'))) as unknown as string
-        const deserializedHEAD = SPMessage.deserializeHEAD(first)
-        if (deserializedHEAD.contractID !== contractID) {
-          controller.error(new Error('chelonia/out/eventsBetween: Mismatched contract ID'))
-          return
-        }
+        const deserializedHEAD = await sbp('chelonia/out/deserializedHEAD', startHash, { contractID })
```
taoeffect: Somewhat confusing API imo.

1. The `contractID` appears to be optional as it's typed `contractID?: string`, yet the implementation always does the comparison, indicating it's mandatory.
2. So if it's mandatory I wouldn't split the parameters like this, but instead move `hash` into the object. Also better either way as it's named parameters like that and named parameters are easier to read at the call site

taoeffect (changes requested):
Great work on this @corrideat ! Minor request re API design:

corrideat:
@taoeffect If you approve, please don't merge it yet, as I'll work on `ChatMain` and there may be other changes yet to be made.

PR #24, opened 2025-11-03
Add `iterKeys` and `keyCount` selectors

corrideat:


src/db.ts
```diff
@@ -112,6 +119,13 @@ const dbPrimitiveSelectors =
         // eslint-disable-next-line require-await
         'chelonia.db/delete': async function (key: string): Promise<boolean> {
           return sbp('okTurtles.data/delete', key)
+        },
+        // eslint-disable-next-line require-await
+        'chelonia.db/iterKeys': async function * () {
+          yield * sbp('okTurtles.data/iterKeys')
+        },
+        'chelonia.db/keyCount': function (): Promise<number> {
+          return Promise.resolve(sbp('okTurtles.data/keyCount'))
```
taoeffect: Sonnet 4.5 notices:

## Minor observations:

- **Type safety**: You might want to add parameter/return types to the first function:
  ```ts
  'chelonia.db/iterKeys': async function * (): AsyncGenerator<YourType> {
    yield * sbp('okTurtles.data/iterKeys')
  }
  ```

- **Consistency**: The `eslint-disable-next-line require-await` comment suggests the linter expects `await` in async functions, but async generators don't require it (the comment is appropriate here).

PR #31, opened 2026-01-04
Add operation hook

corrideat:
Relates to https://github.com/okTurtles/group-income/issues/2988

### DIFF
--- a/src/internals.ts
+++ b/src/internals.ts
@@ -942,22 +942,24 @@ export default sbp('sbp/selectors/register', {
     this: CheloniaContext,
     contractID: string,
     message: SPMessage,
-    state: ChelContractState,
-    atomicIndex?: number
+    state: ChelContractState
   ) {
     if (this.config.skipActionProcessing) return
     const rootState = sbp('chelonia/rootState')
-    const contractName = rootState.contracts[contractID].type || state._vm?.type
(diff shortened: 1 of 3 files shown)


src/internals.ts
```diff
@@ -938,6 +938,32 @@ export default sbp('sbp/selectors/register', {
       targetState._volatile.pendingKeyRequests.filter((pkr) => pkr?.name !== signingKey.name)
     )
   },
+  'chelonia/private/operationHook': function (
```
taoeffect: Seems to make sense for users to configure this via `chelonia/configure`

src/internals.ts
```diff
@@ -938,6 +938,31 @@ export default sbp('sbp/selectors/register', {
...
+    // Check if a hook is defined
+    if (sbp('sbp/selectors/fn', hook)) {
+      // And call it
+      try {
+        sbp(hook, { contractID, message, state, atomicIndex })
+      } catch (e) {
+        console.error(
+          `[chelonia/private/operationHook] Error at operation hook for ${contractID}`,
+          e
+        )
```
taoeffect: Let's also log the hook that failed so the developer knows where to look, so something like this:

```js
        console.error(
          `[${hook}] hook error for message ${messageID} on contract ${contractID}:`,
          e
        )
```

taoeffect (approved):
🎉

Issue #2, opened 2025-05-20
Copy latest changes from Group Income


taoeffect:
### Problem

Recently merged PRs in Group Income made some more changes to Chelonia.

At least this PR has new changes: https://github.com/okTurtles/group-income/pull/2790

### Solution

Copy those changes over.

PR #33, opened 2026-01-09
Key rotation improvements

corrideat:
This relates to https://github.com/okTurtles/group-income/issues/2988

When debugging that issue, I found that sometimes key rotations for foreign keys (which should happen automatically) don't occur. This PR enhances how foreign key rotations are handled by:

1. Correctly evicting keys from `pendingKeyRotations` if a more recent key with the same name is rotated. This prevents lingering keys.
2. When a key rotation needs to be performed, use the most recent key by the same name instead of the literal key ID in `pendingKeyRotations`, which addresses key rotations.

---

This PR also adds `foreignContractIDs` to key shares state, which is used to know whom a key was shared with. That didn't exist before, but it now does and it works in non-adversarial settings. (†)

† The adversarial setting would be claiming that a key is shared with a contract but sending undecryptable or wrong information. (edited) 

---

<a href="https://app.devin.ai/review/okturtles/libcheloniajs/pull/33">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/internals.ts
+++ b/src/internals.ts
@@ -1201,7 +1201,13 @@ export default sbp('sbp/selectors/register', {
               if (!sharedKeyId.foreignContractIDs) {
                 sharedKeyId.foreignContractIDs = [[v.foreignContractID, height]]
               } else {
-                sharedKeyId.foreignContractIDs.push([v.foreignContractID, height])
+                const tuple = sharedKeyId
+                  .foreignContractIDs.find(([id]) => id === v.foreignContractID)
+                if (tuple) {
+                  tuple[2] = height
+                } else {
+                  sharedKeyId.foreignContractIDs.push([v.foreignContractID, height])
+                }
               }
             }
           }
@@ -2397,7 +2403,7 @@ export default sbp('sbp/selectors/register', {
         const [currentRingLevel, currentSigningKeyId, currentKeyIds] = acc
         const ringLevel = Math.min(
           currentRingLevel,
-          contractState._vm?.authorizedKeys?.[keyId]?.ringLevel ?? Number.POSITIVE_INFINITY
+          contractState._vm?.authorizedKeys?.[keyId]?.ringLevel ?? Number.MAX_SAFE_INTEGER
         )
         if (ringLevel >= currentRingLevel) {
           affectedKeyIds.add(keyId)
--- a/dist/cjs/internals.cjs
+++ b/dist/cjs/internals.cjs
@@ -919,7 +919,14 @@ exports.default = (0, sbp_1.default)('sbp/selectors/register', {
                                 sharedKeyId.foreignContractIDs = [[v.foreignContractID, height]];
                             }
                             else {
-                                sharedKeyId.foreignContractIDs.push([v.foreignContractID, height]);
+                                const tuple = sharedKeyId
+                                    .foreignContractIDs.find(([id]) => id === v.foreignContractID);
+                                if (tuple) {
+                                    tuple[2] = height;
+                                }
+                                else {
+                                    sharedKeyId.foreignContractIDs.push([v.foreignContractID, height]);
+                                }
                             }
                         }
                     }
@@ -1873,7 +1880,7 @@ exports.default = (0, sbp_1.default)('sbp/selectors/register', {
             if (!keyId || affectedKeyIds.has(keyId))
                 return acc;
             const [currentRingLevel, currentSigningKeyId, currentKeyIds] = acc;
-            const ringLevel = Math.min(currentRingLevel, contractState._vm?.authorizedKeys?.[keyId]?.ringLevel ?? Number.POSITIVE_INFINITY);
+            const ringLevel = Math.min(currentRingLevel, contractState._vm?.authorizedKeys?.[keyId]?.ringLevel ?? Number.MAX_SAFE_INTEGER);
             if (ringLevel >= currentRingLevel) {
                 affectedKeyIds.add(keyId);
                 currentKeyIds.push(keyId);
--- a/dist/cjs/types.d.cts
+++ b/dist/cjs/types.d.cts
@@ -270,7 +270,7 @@ export type ChelContractState = {
             id: string;
             contractID: string;
             height: number;
-            foreignContractIDs?: [string, number][];
+            foreignContractIDs?: ([contractID: string, firstShareHeight: number] | [contractID: string, firstShareHeight: number, lastShareHeight: number])[];
             keyRequestHash?: string;
             keyRequestHeight?: number;
         }[];
--- a/dist/esm/internals.mjs
+++ b/dist/esm/internals.mjs
@@ -884,7 +884,14 @@ export default sbp('sbp/selectors/register', {
                                 sharedKeyId.foreignContractIDs = [[v.foreignContractID, height]];
                             }
                             else {
-                                sharedKeyId.foreignContractIDs.push([v.foreignContractID, height]);
+                                const tuple = sharedKeyId
+                                    .foreignContractIDs.find(([id]) => id === v.foreignContractID);
+                                if (tuple) {
+                                    tuple[2] = height;
+                                }
+                                else {
+                                    sharedKeyId.foreignContractIDs.push([v.foreignContractID, height]);
+                                }
                             }
                         }
                     }
@@ -1838,7 +1845,7 @@ export default sbp('sbp/selectors/register', {
             if (!keyId || affectedKeyIds.has(keyId))
                 return acc;
             const [currentRingLevel, currentSigningKeyId, currentKeyIds] = acc;
-            const ringLevel = Math.min(currentRingLevel, contractState._vm?.authorizedKeys?.[keyId]?.ringLevel ?? Number.POSITIVE_INFINITY);
+            const ringLevel = Math.min(currentRingLevel, contractState._vm?.authorizedKeys?.[keyId]?.ringLevel ?? Number.MAX_SAFE_INTEGER);
             if (ringLevel >= currentRingLevel) {
                 affectedKeyIds.add(keyId);
                 currentKeyIds.push(keyId);
--- a/dist/esm/types.d.mts
+++ b/dist/esm/types.d.mts
@@ -270,7 +270,7 @@ export type ChelContractState = {
             id: string;
             contractID: string;
             height: number;
-            foreignContractIDs?: [string, number][];
+            foreignContractIDs?: ([contractID: string, firstShareHeight: number] | [contractID: string, firstShareHeight: number, lastShareHeight: number])[];
             keyRequestHash?: string;
             keyRequestHeight?: number;
         }[];
--- a/src/types.ts
+++ b/src/types.ts
@@ -288,7 +288,10 @@ export type ChelContractState = {
       contractID: string;
       height: number;
       // List of contract IDs the key share is addressed to
-      foreignContractIDs?: [string, number][];
+      foreignContractIDs?: (
+        | [contractID: string, firstShareHeight: number]
+        | [contractID: string, firstShareHeight: number, lastShareHeight: number]
+      )[];
       keyRequestHash?: string;
       keyRequestHeight?: number;
     }[];


taoeffect (changes requested):
Prelim machine review, ran it through two:

### Opus 4.5

<details>


## Critical Bug: Missing Null Check

In the `chelonia/private/deleteOrRotateRevokedKeys` function, there's a potential `TypeError` when accessing a key that may not exist in `activeForeignKeyIds`:

```typescript
const activeKeyId = activeForeignKeyIds[pkrKey.foreignKey]
const key = contractState._vm.authorizedKeys[activeKeyId]
if (affectedKeyIds.has(key.id)) return acc
```

If `pkrKey.foreignKey` doesn't have a corresponding active key (e.g., because all keys with that foreign key have been revoked), `activeKeyId` will be `undefined`, making `key` undefined, and `key.id` will throw a `TypeError`.

**Suggested fix:**

```typescript
const activeKeyId = activeForeignKeyIds[pkrKey.foreignKey]
if (!activeKeyId) return acc  // No active key exists for this foreign key
const key = contractState._vm.authorizedKeys[activeKeyId]
if (affectedKeyIds.has(key.id)) return acc
```

## Potential Issue: Null Safety in `deleteKeyHelper`

The helper function assumes `state._volatile` and `state._volatile.pendingKeyRevocations` exist:

```typescript
for (const id of allIdsForNames[name]) {
  if (has(state._volatile!.pendingKeyRevocations, id)) {
    delete state._volatile!.pendingKeyRevocations![id]
  }
}
```

While the call sites currently initialize these properties beforehand, the function would be more robust with a guard:

```typescript
const deleteKeyHelper = (state: ChelContractState, height: number, keyIds: string[]) => {
  const allIdsForNames = Object.values(state._vm.authorizedKeys)
    .reduce<Record<string, string[]>>((acc, { id, name }) => {
      if (!acc[name]) {
        acc[name] = [id]
      } else {
        acc[name].push(id)
      }
      return acc
    }, Object.create(null))
  for (const keyId of keyIds) {
    const name = state._vm.authorizedKeys[keyId].name
    // Clear pending revocations for all keys with the same name
    // to handle key rotation scenarios where multiple keys exist
    for (const id of allIdsForNames[name]) {
      if (state._volatile?.pendingKeyRevocations && has(state._volatile.pendingKeyRevocations, id)) {
        delete state._volatile.pendingKeyRevocations[id]
      }
    }

    state._vm.authorizedKeys[keyId]._notAfterHeight = height
  }
}
```

## Improvement: Type Annotation for Reduce Accumulator

The reduce accumulator in `deleteKeyHelper` lacks explicit typing:

```typescript
.reduce((acc, { id, name }) => {
  // ...
}, Object.create(null))
```

**Suggested improvement:**

```typescript
.reduce<Record<string, string[]>>((acc, { id, name }) => {
  // ...
}, Object.create(null))
```

## Behavioral Change Clarification

The new `deleteKeyHelper` clears pending revocations for **all keys with the same name**, not just the specific key being deleted. The comment explains this is "to handle key rotation scenarios where multiple keys exist." 

This is a significant behavioral change from the previous code which only cleared the specific `keyId`. Please confirm this broadened scope is intentional and won't cause unintended side effects when multiple keys share the same name but should be handled independently.
</details>

### Gemini 3 Pro

<details>

There are **two critical bugs** introduced in the `chelonia/private/deleteOrRotateRevokedKeys` selector. One causes a runtime crash (DoS), and the other causes logical errors where keys may fail to update if multiple local keys reference the same foreign key.

### 1. Critical: Runtime Crash (DoS) in `deleteOrRotateRevokedKeys`

In the `reduce` function within `deleteOrRotateRevokedKeys`, the code attempts to resolve an active key ID from a foreign key string. If the foreign key is not found in `activeForeignKeyIds` (which can happen if the key was deleted locally, or due to the overwrite bug detailed in point #2), `activeKeyId` becomes `undefined`.

Subsequently, accessing `contractState._vm.authorizedKeys[undefined]` returns `undefined`. Trying to access `.id` on that `undefined` result throws a TypeError, crashing the message processing.

**Problematic Lines:**
```typescript
const activeKeyId = activeForeignKeyIds[pkrKey.foreignKey]
const key = contractState._vm.authorizedKeys[activeKeyId]
if (affectedKeyIds.has(key.id)) return acc
```

**Recommended Fix:**
You must check if `activeKeyId` (and consequently `key`) exists before accessing properties on it.

```typescript
const activeKeyId = activeForeignKeyIds[pkrKey.foreignKey]
// FIX: Check if activeKeyId exists
if (!activeKeyId) return acc

const key = contractState._vm.authorizedKeys[activeKeyId]
// FIX: Ensure key lookup succeeded (sanity check)
if (!key) return acc 

if (affectedKeyIds.has(key.id)) return acc
```

### 2. Critical: Logic Error in `activeForeignKeyIds` Map Construction

The construction of `activeForeignKeyIds` uses `Object.fromEntries` with the foreign key URL as the key. If a contract has **multiple** local keys pointing to the **same** foreign key URL (e.g., different permissions or ring levels for the same foreign identity), `Object.fromEntries` will overwrite previous entries. Only the last processed key will be stored.

This results in the `reduce` loop failing to update/rotate the keys that were overwritten in the map.

**Problematic Lines:**
```typescript
const activeForeignKeyIds = Object.fromEntries(
  Object.values(contractState._vm.authorizedKeys)
    .filter(({ foreignKey, _notAfterHeight }) =>
      foreignKey != null && _notAfterHeight == null
    )
    .map(({ foreignKey, id }) => [foreignKey!, id])
)
```

**Recommended Fix:**
Since `deleteOrRotateRevokedKeys` is trying to find the *current* active key that corresponds to a revoked key, and multiple keys might share a foreign key string, this logic is brittle.

However, to fix the immediate overwrite issue without refactoring the entire logic, you should map the Foreign Key string to an **Array** of IDs, and iterate over that array in the reducer.

**Revised Code Suggestion (Fixing #1 and #2):**

```typescript
// 1. Create a map of ForeignKey -> Array<LocalKeyID>
const activeForeignKeyIds = Object.values(contractState._vm.authorizedKeys)
  .filter(({ foreignKey, _notAfterHeight }) =>
    foreignKey != null && _notAfterHeight == null
  )
  .reduce((acc, { foreignKey, id }) => {
    const fk = foreignKey!
    if (!acc[fk]) acc[fk] = []
    acc[fk].push(id)
    return acc
  }, Object.create(null) as Record<string, string[]>)

// ... inside the reducer ...

const [, keyUpdateSigningKeyId, keyUpdateArgs] = keysToUpdate.reduce(
  (acc, keyId) => {
    const pkrKey = contractState._vm?.authorizedKeys?.[keyId]
    if (!pkrKey || !pkrKey.foreignKey) return acc
    
    // 2. Retrieve the list of active IDs for this foreign key
    const activeKeyIds = activeForeignKeyIds[pkrKey.foreignKey]
    if (!activeKeyIds || activeKeyIds.length === 0) return acc

    // 3. Iterate over all potential active keys (handling the 1-to-many case)
    let currentAcc = acc
    
    for (const activeKeyId of activeKeyIds) {
      const key = contractState._vm.authorizedKeys[activeKeyId]
      if (!key) continue // Safety check

      // Prevent processing the same key twice
      if (affectedKeyIds.has(key.id)) continue

      // ... existing logic regarding ringLevel and finding signingKeyId ...
      // Note: You will need to adapt the accumulator logic here to handle 
      // potentially adding multiple updates in one pass of the reducer, 
      // or ensure the logic selects the specific correct key based on `name`.
      
      // OPTIMIZATION: Usually, a key rotation replaces a key with the SAME name.
      // It is safer to filter by name here:
      if (key.name !== pkrKey.name) continue

      // ... proceed with existing logic ...
    }
    return currentAcc
  },
  // ... initial value ...
)
```

### 3. Improvement: Type Safety in `deleteKeyHelper`

In `deleteKeyHelper`, you access `state._volatile!` using the non-null assertion operator. While the callers (`OP_KEY_DEL`, `OP_KEY_UPDATE`) currently ensure `_volatile` exists, relying on the caller for null safety in a helper function can lead to future bugs if the helper is reused elsewhere.

**Problematic Lines:**
```typescript
if (has(state._volatile!.pendingKeyRevocations, id)) {
  delete state._volatile!.pendingKeyRevocations![id]
}
```

**Improvement:**
Use optional chaining.

```typescript
const deleteKeyHelper = (state: ChelContractState, height: number, keyIds: string[]) => {
  // ... existing grouping logic ...

  for (const keyId of keyIds) {
    const key = state._vm.authorizedKeys[keyId]
    if (!key) continue // Safety check in case keyId is invalid

    const name = key.name
    
    // Safe access
    if (state._volatile?.pendingKeyRevocations) {
      for (const id of allIdsForNames[name] || []) {
        if (has(state._volatile.pendingKeyRevocations, id)) {
           delete state._volatile.pendingKeyRevocations[id]
        }
      }
    }

    state._vm.authorizedKeys[keyId]._notAfterHeight = height
  }
}
```

### 4. Logic Check: `deleteKeyHelper` grouping

In `deleteKeyHelper`:
```typescript
const allIdsForNames = Object.values(state._vm.authorizedKeys)
// ...
for (const id of allIdsForNames[name]) {
```
If `keyIds` contains a key ID that does not exist in `state._vm.authorizedKeys`, the line `const name = state._vm.authorizedKeys[keyId].name` will throw.

While `OP_KEY_DEL` filters keys before calling this helper, `OP_KEY_UPDATE` passes `keysToDelete` derived from `validateKeyUpdatePermissions`. If that validation function returns an ID that isn't in `authorizedKeys` (unlikely, but possible during race conditions or state desyncs), this helper will crash.

**Suggestion:** Add a simple existence check inside the loop in `deleteKeyHelper` (as shown in the code block in point #3 above).
</details>

corrideat:
This issue:

```js
const activeKeyId = activeForeignKeyIds[pkrKey.foreignKey]
// [suggested] FIX: Check if activeKeyId exists
// ...
const key = contractState._vm.authorizedKeys[activeKeyId]
if (affectedKeyIds.has(key.id)) return acc
```

Isn't really possible due to how `activeForeignKeyIds` is constructed.

> Critical: Logic Error in activeForeignKeyIds Map Construction

It definitely is weird that a contract might have a foreign key defined as a key multiple time, and doing this will break `authorizedKeys`, which is keyed by key ID. Since things already break when this is attempted, then perhaps we shouldn't worry too much about it.

> Potential Issue: Null Safety in deleteKeyHelper
> > While the call sites currently initialize these properties beforehand

Dismissing adding a guard since it's redundant (a DRY argument can be made for moving the `_volatile` logic, but, personally, I prefer it as it currently is since it's clear and explicit.

> Type Annotation for Reduce Accumulator

This would IMHO be more verbose and confusing than the current implicit approach.

>  Improvement: Type Safety in deleteKeyHelper
> > [...] While the callers [...] currently ensure _volatile exists [...]

Dismissing

>  Logic Check: deleteKeyHelper grouping

This check is redundant because all keys are supposed to exist, and their non-existence means that there's a bug elsewhere.

src/chelonia.ts
```diff
@@ -310,6 +314,7 @@ export default sbp('sbp/selectors/register', {
       // Similarly, future events will not be reingested and will throw
       // with ChelErrorDBBadPreviousHEAD
       strictOrdering: false,
+      saveMessageMetadata: false,
```
taoeffect: Likewise a 1-line comment explaining this parameter (what's saved where?)

taoeffect (changes requested):
Partial review!

corrideat:
Re. the pending key rotations race condition on foreign keys (see https://github.com/okTurtles/libcheloniajs/issues/45):

This is a bug we _currently_ have, so not fixing it isn't any worse than what we currently do.

In terms of impact, it means that if a contract is being synced from scratch it may not update foreign keys that need to be rotated the way one would expect. In other words, keys that already are out of sync will continue to be out of sync after syncing from scratch.

This shouldn't _break_ things (and if it does, that's another bug), but it's a mistake because we're relying on keys that the authoritative source (the owner contract) no longer recognises as valid.

A concrete example: someone leaves a group (causing a key rotation) and the `#general` chatroom for some reason doesn't reflect this rotation, meaning the former member can still read and sign messages. Syncing from scratch may not fix this situation the way one would expect.

taoeffect:
Using the new prompt and the cheaper Copilot plan, here's another Opus 4.6 review:

## Opus 4.6 (with tools, improved prompt)

<details>

Now I have enough context for a thorough review.

---

# Code Review

## 🔴 High Importance

### 1. `keyIds.length === 0` sends a misleading success response with no keys

- `internals.ts:2647-2648`

When the response hook returns an empty `keyIds` array, the code returns `[null, skipInviteAccounting]`. This flows into the second `.then()` where `keySharePayload` is `null`, so no `OP_KEY_SHARE` is sent—but the code still sends an `OP_KEY_REQUEST_SEEN` with `{ success: true }` and `keyShareHash: undefined`.

```typescript
} else if (keyIds.length === 0) {
  return [null, skipInviteAccounting]
}
```

Then at line 2718:
```typescript
const innerPayload = { keyShareHash: msg?.hash(), success: true }
```

This means: "I successfully shared keys (none of them)" which is semantically wrong. The requester will think the key share succeeded but will have received zero keys. This should either be treated as `success: false` or should not send a response at all.

**Suggestion:** If no keys need to be shared, either return early (like the `!Array.isArray(keyIds)` case) or explicitly send `success: false`:
```typescript
} else if (keyIds.length === 0) {
  return
}
```

---

### 2. Invite expiry check in `respondToKeyRequest` ignores `requestedSkipInviteAccounting`

- `internals.ts:2582-2593`

When `request == null || request === '*'`, the responder unconditionally checks invite expiry, even when `requestedSkipInviteAccounting` is `true`. This is inconsistent with the `OP_KEY_REQUEST` handler (line 1487-1488) which skips all invite checks when `skipInviteAccounting` is true.

```typescript
if (request == null || request === '*') {
  if (contractState._vm?.invites?.[inviteId]?.expires != null) {
    if (contractState._vm.invites[inviteId].expires < Date.now()) {
      // ... return (silently reject)
    }
  }
```

If a V2 request comes in with `skipInviteAccounting: true` and `request: '*'`, the OP_KEY_REQUEST handler lets it through, but the responder rejects it at this point if the invite has expired.

**Suggestion:** Guard the expiry check with `requestedSkipInviteAccounting`:
```typescript
if (request == null || request === '*') {
  if (!requestedSkipInviteAccounting && contractState._vm?.invites?.[inviteId]?.expires != null) {
    if (contractState._vm.invites[inviteId].expires < Date.now()) {
      // ...
    }
  }
```

---

### 3. `pendingKeyshares` writes 7-element tuples but type allows 4-element—yet 4-element tuples are never written

- `internals.ts:1557-1569`
- `types.ts:295-312`

The code now always writes either a 3-element or a 7-element tuple. The 4-element variant in the type definition is never produced by the current code. However, all consumer checks guard on `entry.length !== 4 && entry.length !== 7` (lines 1596, 2483, 2513), so they accept the 4-element form. When they destructure a 4-element entry, `request`, `manifestHash`, and `requestedSkipInviteAccounting` will be `undefined`.

This works for backward compatibility (processing old pendingKeyshares entries), but the destructuring at `internals.ts:2517-2525` will produce `requestedSkipInviteAccounting = undefined` for 4-element entries, while the mismatch check at line 2693 coerces both sides with `!!`:

```typescript
if (!!requestedSkipInviteAccounting !== !!skipInviteAccounting) {
```

For a 4-element entry, `requestedSkipInviteAccounting` is `undefined` and `skipInviteAccounting` will also likely be `undefined` (since `request` is `undefined` → takes `request == null` branch → `skipInviteAccounting` stays `undefined`). So this works, but it's fragile—a 4-element entry with a `*` request from old state could subtly break if the hook logic changes. Worth a comment noting the backward-compat assumption.

---

## 🟡 Medium Importance

### 4. `OP_KEY_UPDATE` handler does spread-merge that can drop `_notBeforeHeight` and other internal fields

- `internals.ts:1792-1799`

When updating an existing key, the code spreads only select fields over the existing key:

```typescript
state._vm.authorizedKeys[key.id] = {
  ...state._vm.authorizedKeys[key.id],
  ...(key.purpose ? { purpose: key.purpose } : {}),
  ...(key.permissions ? { permissions: key.permissions } : {}),
  ...(key.allowedActions ? { allowedActions: key.allowedActions } : {}),
  ...(key.meta ? { meta: key.meta } : {})
}
```

This replaces the entire object reference (not an in-place merge). It preserves `_notBeforeHeight`, `id`, `name`, `data`, `foreignKey`, `ringLevel`, etc. from the spread of the existing entry. However, `meta` is fully replaced, not deep-merged. If the existing key has `meta.private` and the update only sends `meta.keyRequest`, the `private` field is lost. Verify this is intentional.

---

### 5. `operationHook` can double-fire if an inner OP_ATOMIC op type matches `OP_ATOMIC`

- `internals.ts:1005-1014`

The `callHook` for `message.opType()` is unconditional. If `OP_ATOMIC` itself has a registered hook, it will fire in addition to the per-inner-op hooks. More importantly, if an inner op's type string ever equals the `OP_ATOMIC` type string, the hook would fire twice—once with `atomic=true`, once with `atomic=undefined`.

```typescript
if (message.opType() === SPMessage.OP_ATOMIC) {
  const opsSet = new Set()
  for (const [op] of message.opValue() as SPOpAtomic) {
    if (opsSet.has(op)) continue
    opsSet.add(op)
    callHook(op, true)
  }
}
callHook(message.opType())  // Always fires, even for OP_ATOMIC
```

**Suggestion:** If the intent is that `OP_ATOMIC` only decomposes into its inner ops, wrap line 1014 in an `else`:
```typescript
if (message.opType() === SPMessage.OP_ATOMIC) {
  // ...per-inner-op hooks...
} else {
  callHook(message.opType())
}
```
Or if you want both (a per-op hook AND a batch hook), add `OP_ATOMIC` to `opsSet` to prevent the exact-match double-call:
```typescript
opsSet.add(SPMessage.OP_ATOMIC)  // prevent double-fire
```

---

### 6. `deleteOrRotateRevokedKeys` uses `pkrKey.foreignKey` to look up active key, but multiple active keys could share the same `foreignKey` URI

- `internals.ts:2304-2310`

```typescript
const activeForeignKeyIds = Object.fromEntries(
  Object.values(contractState._vm.authorizedKeys)
    .filter(({ foreignKey, _notAfterHeight }) =>
      foreignKey != null && _notAfterHeight == null
    )
    .map(({ foreignKey, id }) => [foreignKey!, id])
)
```

If two active keys share the same `foreignKey` URI (unlikely but not prevented by the type system), `Object.fromEntries` will keep only the last one, silently dropping the other. This could cause a key to never get rotated or deleted.

---

### 7. `krsEncryption` changed from checking `_private` on signing key to using `keyShareEncryption`

- `internals.ts:2527`

Old code:
```typescript
const krsEncryption = !!contractState._vm.authorizedKeys?.[signingKeyId]?._private
```

New code:
```typescript
const krsEncryption = keyShareEncryption
```

`keyShareEncryption` comes from `encryptedRequest` (index 0 of the pendingKeyshares entry), which tracks whether the *incoming request* was encrypted. The old code checked whether the *responder's signing key* has a `_private` field. These are semantically different checks. If the request was encrypted but the responder's key is not private (or vice versa), the behavior changes. Verify this is the intended semantic shift.

---

### 8. `foreignContractIDs` in `sharedKeyIds` grows unboundedly

- `internals.ts:1218-1225`

Every time a key share comes in for the same key ID from a different (or same) foreign contract, the array gets a new entry appended. There's no deduplication check for `foreignContractID`:

```typescript
} else if (v.foreignContractID) {
  if (!sharedKeyId.foreignContractIDs) {
    sharedKeyId.foreignContractIDs = [[v.foreignContractID, height]]
  } else {
    sharedKeyId.foreignContractIDs.push([v.foreignContractID, height])
  }
}
```

If the same foreign contract ID sends multiple key shares for the same key, duplicates accumulate. Consider deduplicating.

---

## ⚪ Minor Importance

### 9. Typo in comment: "optionl"

- `internals.ts:995`

```typescript
// hooks is optionl (for example, they aren't run on the server)
```

Should be "optional".

---

### 10. `ProtoSPOpKeyRequestInnerV2` is identical to `ProtoSPOpKeyRequestV1`

- `SPMessage.ts:87-95` vs `SPMessage.ts:96-104`

Both types have the exact same shape. If this is intentional for future divergence, a comment would help. Otherwise, consider using a single type with an alias.

---

### 11. `chelonia/out/keyRequest` has a TODO noting `keyRequestResponseId` breaks `pendingKeyRequests`

- `chelonia.ts:2053-2054`

```typescript
// TODO: This won't work if re-using a request key using keyRequestResponseId
```

And again at line 2058-2059. This is a known gap—if `keyRequestResponseId` is provided, the `havePendingKeyRequest` check won't find it, potentially leading to duplicate key requests.

---

### 12. Variable shadowing of `data` in `OP_KEY_REQUEST_SEEN` handler

- `internals.ts:1611`

```typescript
if (data.encryptionKeyId == null && has(v, 'innerData')) {
  const data = config.unwrapMaybeEncryptedData(...)  // shadows outer `data`
  inner = data?.data
}
```

The inner `const data` shadows the outer `data` from line 1587. This works but is confusing and can lead to maintenance bugs. Consider renaming to `innerDataResult` or similar.

---

### 13. Line length exceeds 100 chars in several places

Per the ESLint config (max 100 chars for code), several new lines exceed this:
- `internals.ts:2694` — the `console.error` string
- `chelonia.ts:2064` — the `throw new Error` string
- `chelonia.ts:1400` — the function signature

These will likely trigger lint warnings.

</details>

src/chelonia.ts
```diff
@@ -721,21 +735,30 @@ export default sbp('sbp/selectors/register', {
   },
   'chelonia/contract/setPendingKeyRevocation': function (
     this: CheloniaContext,
-    contractID: string,
-    names: string[]
+    contractIDOrState: string | ChelContractState,
```
taoeffect: Could you please add a comment above this selector explaining:

- What this is for (i.e. what sorts of situations this is useful for, give an example)
- When the user is supposed to call it
- Mention if there are any other selectors the user should call along with it, i.e. if this selector is to be used in conjunction with any other

src/chelonia.ts
```diff
@@ -1995,89 +2027,112 @@ export default sbp('sbp/selectors/register', {
...
       const havePendingKeyRequest =
-        Object.values(originatingState._vm.authorizedKeys).findIndex((k: ChelContractKey) => {
+        Object.values(originatingState._vm.authorizedKeys).some((k: ChelContractKey) => {
           return (
             k._notAfterHeight == null &&
             k.meta?.keyRequest?.contractID === contractID &&
-            state?._volatile?.pendingKeyRequests?.some((pkr) => pkr.name === k.name)
+            state?._volatile?.pendingKeyRequests?.some(
+              (pkr) => pkr.name === k.name && pkr.reference === reference
+            )
```
taoeffect: What if the `reference` is not included in the params, but the `pendingKeyRequests` only contain a pkr with a reference?

Don't we still want to match in that case?

src/chelonia.ts
```diff
@@ -1995,89 +2038,112 @@ export default sbp('sbp/selectors/register', {
...
-        [SPMessage.OP_KEY_ADD],
-        ['sig']
-      )
-      if (!signingKeyId) {
-        throw new ChelErrorUnexpected(
-          `Unable to send key request. Originating contract is missing a key with OP_KEY_ADD permission. contractID=${contractID} originatingContractID=${originatingContractID}`
+      if (keyRequestResponseId) {
+        // Use an exisiting key (i.e., avoid sending OP_KEY_ADD)
+        // TODO: This option currently breaks `pendingKeyRequests` due to this
+        // being set on `OP_KEY_ADD`.
```
taoeffect: Is this something that should be fixed in this PR?

If not, can you also add a comment saying what the fix should be?

src/chelonia.ts
```diff
@@ -213,6 +214,12 @@ export type ChelKeyRequestParams = {
   // Arbitrary data the requester can use as reference (e.g., the hash
   // of the user-initiated action that triggered this key request)
   reference?: string;
+  // Contract-defined string describing which keys are being requested
+  request?: string;
```
taoeffect: Could you give an example in the comment as to what this looks like?

corrideat:
I've finished going over the LLM reviews and should have addressed most things there. They all seemed about as equally effective (and there was a lot of overlap).

Most of the issues they pointed out were addressed with a comment (there was no issue other than clarity), were minor robustness issues (e.g., adding checks for potentially `undefined` fields that can't be undefined based on how things are called) or were issues but still minor (e.g., adding a guard to prevent duplicate sharing).

src/SPMessage.ts
```diff
@@ -92,13 +91,44 @@ export type ProtoSPOpKeyRequest = {
   }>;
   request: string;
 };
-export type SPOpKeyRequest = ProtoSPOpKeyRequest | EncryptedData<ProtoSPOpKeyRequest>;
-export type ProtoSPOpKeyRequestSeen = {
+// Same data and structure as ProtoSPOpKeyRequestV1. Difference between V1 and
+// V2 is an unencrypted wrapper
```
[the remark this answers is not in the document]
taoeffect: Could you add that explanation to the code then?

src/chelonia.ts
```diff
@@ -721,21 +735,30 @@ export default sbp('sbp/selectors/register', {
...
-    const state = rootState[contractID]
+    let state: ChelContractState
+    let contractID: string | undefined
+    if (typeof contractIDOrState === 'string') {
+      const rootState = sbp(this.config.stateSelector)
+      contractID = contractIDOrState
+      state = rootState[contractIDOrState]
+    } else {
+      state = contractIDOrState
+    }
```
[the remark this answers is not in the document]
taoeffect: Yeah, go ahead and create an issue for it 👍

src/internals.ts
```diff
@@ -1690,17 +1800,52 @@ export default sbp('sbp/selectors/register', {
...
             key._notBeforeHeight = height
             state._vm.authorizedKeys[key.id] = cloneDeep(key)
+          } else {
+            state._vm.authorizedKeys[key.id] = {
+              ...state._vm.authorizedKeys[key.id],
+              ...(key.purpose ? { purpose: key.purpose } : {}),
+              ...(key.permissions ? { permissions: key.permissions } : {}),
+              ...(key.allowedActions ? { allowedActions: key.allowedActions } : {}),
+              ...(key.meta ? { meta: key.meta } : {})
+            }
```
taoeffect: This is kinda ugly tbh... Not saying you need to change it, but if you find a cleaner way of doing this that would be preferable

taoeffect:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/22202235480)._

taoeffect:
/review_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/22202809987)._

taoeffect:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/22203605936)._

taoeffect:
/crush_fast (*oops, I forgot that crush has a bug that limits the GLM-5 context window to half its size. Fixed! re-running)

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/22210528230)._

taoeffect:
/crush_fast {{openrouter, google/gemini-3.1-pro-preview, google/gemini-3-flash-preview}} - cancelling the previous run since a commit was pushed

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/22283191627)._

PR #43, opened 2026-02-07
Add tests for permission escalation

### LINKED ISSUE #40: Verify no key permission esclation exists
### Problem

Verify that, in all cases, `OP_KEY_ADD` and `OP_KEY_UPDATE` can never be used to indirectly escalate their own permissions, for example by having K1 (permissions `OP_KEY_ADD`) used to add a key K2 with more permissions than K1, for example `*`. The same goes for `OP_KEY_UPDATE`.

### Solution

When `OP_KEY_ADD` or `OP_KEY_UPDATE` are used, the following must hold:

1. A key can only be changed by a key with a ring level equal or lower
2. No permissions can be gained from a key though other key operations. For example, a key without `OP_ACTION_ENCRYPTED` cannot add a key with this permission, give it to itself or give it to another key.

corrideat:
Closes #40

---

<a href="https://app.devin.ai/review/okturtles/libcheloniajs/pull/43" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,3 +1,4 @@
 import './encryptedData.test.js'
 import './persistent-actions.test.js'
 import './pubsub/index.test.js'
+import './utils.test.js'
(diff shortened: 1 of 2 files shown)


taoeffect (approved):
👍

PR #44, opened 2026-02-08
Implement kd#self

### LINKED ISSUE #41: Implement contexts and `kd#self` permission
### Problem

Invite creators should be able to delete invites they themselves create, but not other invites.

Invite 'managers' (those with such a permission) should be able to delete any invites.

### Solution

To make this work, we add a field to OP_KEY_ADD called context. Where permissionsContext is an array of strings of MAX_LEN.

Every group member has a key with the following permissions:

* `ka`
* `kd#self` - New special permission that indicates keys created by this key can be deleted by this key

And a `permissionsContext` of `["invites"]`. And a `ringLevel` of 5.

This lets any user create invites that they can delete.

### Revoke other invites ("Revoke Invite permission")

Needs to be able to delete invites, but not other keys. So, this means we need to be able to identify invites.

Why don’t we just rely on `ringLevel` alone? Well, because if we did then that would actually complicate things, because this key would be able to delete all keys in all contexts of equal or greater ring level. And we just want this key to be able to delete all INVITES only, not all keys of equal or greater `ringLevel`.

We solve this with a context that applies to all invites, let’s call it the `"invites"` context and we make this permission a lower `ringLevel`.

### Contexts vs Ring Levels

Think of the permissions system as a cone. The cone shape is the context. And how far down the cone you go is the `ringLevel`.

Or another way of thinking of it: all of the keys can be thought of as living in a pie. A context is the slice of the pie. And the `ringLevel` is how much of that slice you’re affecting.

corrideat:
Partially closes #41 (`permissionsContext` not included, planned for later)

---

<a href="https://app.devin.ai/review/okturtles/libcheloniajs/pull/44" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>

### DIFF
--- a/dist/cjs/utils.cjs
+++ b/dist/cjs/utils.cjs
@@ -122,8 +122,8 @@ const validateKeyPermissions = (msg, config, state, signingKeyId, opT, opV) => {
         !signingKey.purpose.includes('sig') ||
         (signingKey.permissions !== '*' &&
             (!Array.isArray(signingKey.permissions) || (!signingKey.permissions.includes(opT) &&
-                (opT === SPMessage_js_1.SPMessage.OP_KEY_DEL &&
-                    !signingKey.permissions.includes(`${opT}#self`)))))) {
+                !(opT === SPMessage_js_1.SPMessage.OP_KEY_DEL &&
+                    signingKey.permissions.includes(`${opT}#self`)))))) {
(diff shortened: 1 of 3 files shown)


taoeffect:
/crush_fast {{openrouter, google/gemini-3.1-pro-preview, google/gemini-3-flash-preview}}

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/22267496556)._

PR #72, opened 2026-05-07
Add new selector for deserializing messages

### LINKED ISSUE #71: Decryption / deserialization selector
### Problem

This is a sub-issue of <https://github.com/okTurtles/chel/issues/113>.

As of now, there's no easy way to decrypt messages. It's possible to call `'chelonia/in/processMessage'`, which with a high enough debug level could result in the same information, indirectly, but there's no standalone way to decrypt messages.

This is because we don't expose `SPMessage` and we also don't expose decryption primitives (although the latter wouldn't be as useful, since there are some things `SPMessage` does in order to decrypt a message, such as pass on the message head as additional data, that would need to be done manually).

### Solution

Since we don't seem to have many issues with the decryption primitives themselves, the first order of business, also useful for 113, would be creating a new selector to deserialize (including decryption) an `SPMessage`. This is very similar to `'chelonia/in/processMessage'`, except that the message wouldn't be processed into the state (as we do this, we might also rewrite `'chelonia/in/processMessage'` to use this new selector)

We should also consider exposing the crypto primitives in the SW in Group Income (as we do with `sbp`) for convenience.

corrideat:
Closes #71

### DIFF
--- a/dist/cjs/chelonia.cjs
+++ b/dist/cjs/chelonia.cjs
@@ -1025,6 +1025,42 @@ exports.default = (0, sbp_1.default)('sbp/selectors/register', {
             signingKeyId: (0, utils_js_1.findSuitableSecretKeyId)(contractState, [SPMessage_js_1.SPMessage.OP_KEY_DEL], ['sig'])
         });
     },
+    // Helper function to deserialize an `SPMessage` from a raw string. This
+    // function invokes `SPMessage.deserialize` with the correct internal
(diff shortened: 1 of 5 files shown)


corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/25573730970)._

taoeffect (approved):
LGTM!

PR #75, opened 2026-05-16
Implement journal

### LINKED ISSUE #74: Implement a journaling mechanism for contracts
### Problem

Debugging certain types of failures in a Chelonia app can be quite difficult sometimes, especially if the failure is a result of some complicated code deep inside the internals of Chelonia. Having the application logs sometimes is not enough to track down after-the-fact what went wrong. Having the events themselves is also not always useful because sometimes one device will process them without issue, while another device mysteriously failed.

What's missing is the ability to see what the Chelonia state was on a client at each particular point in time, before and after an event is processed.

### Solution

What's needed is a compact diff-based representation of state changes. This is called ["journal"](https://en.wikipedia.org/wiki/Journal_(computing)).

Using something akin to [jsonpatch](https://jsonpatch.com/), create a journal that keeps between X and 2X number of messages.

The start of the journal would be a state snapshot (with redactions applied, see below), followed by an action and the diff. Once it reached X messages another snapshot would be saved at that point. Following that another X patches, and once it reached a total of 2X events the first half of the log would be dropped, leaving X messages with a snapshot at the start. This method avoids having to recompute the snapshot at the start of the log on every message.

Snapshotting would be enabled on select `contractIDs` (or globally on all contracts) via `chelonia/configure`, which would also take in an argument that would specify key-paths in the state object to censor before writing them to the journal (to protect secret keys and possibly other information). 

The redaction would take in a key path (e.g. `"key1.key2.key3"`) to redact fields deep inside the nested state object, plus a simple transformation function that would take in that key's value and return a redacted value. For example, to censor secret keys while giving the ability to distinguish them, the transformation function could apply a cryptographic hash and return the first 8 characters of that hash.

corrideat:
Closes #74 (assisted by Opus 4.7)


src/journal.ts
```diff
@@ -0,0 +1,719 @@
...
+  }
+  if (after === undefined) {
+    // RFC 6902 does not define `remove` at the document root, so we emit
+    // `replace` with `null` at root instead. At non-root positions, a
+    // proper `remove` is well-defined and that's what we use.
+    if (segments.length === 0) {
+      out.push({ op: 'replace', path, value: null })
+    } else {
+      out.push({ op: 'remove', path })
+    }
```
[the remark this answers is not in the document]
corrideat: Since contract state elements aren't supposed to have `undefined` values (these will be dropped or converted to `null` when going through `JSON.stringify`), this is fine.

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/26000626833)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/26001137203)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/26002127920)._

corrideat:
<details>
<summary>Review by Opus 4.7</summary>

# Code Review

**Base**: `main`
**Head**: `74-implement-a-journaling-mechanism-for-contracts` (HEAD `5a27466`)
**Date**: 2026-05-18

Scope: per-contract event journal — `src/journal.ts`, integration into
`handleEvent.applyProcessResult` in `src/internals.ts`, config wiring in
`src/chelonia.ts`, new exports/types, and tests.

Overall, the implementation is careful (own-property walks, defended
`__proto__`/`constructor` writes, RFC 6902 strict-subset semantics,
"recorder MUST NOT throw" contract, snapshot/trim invariants, resync
vs. duplicate-arrival distinction). The issues below are mostly about
the runtime-reconfigure surface diverging from what `AGENTS.md`
promises, plus a few footguns around persistence and lossy clones.

---

## 1. 🔴 Field-level `null` is silently accepted instead of rejected as documented

- [x] Addressed
- [ ] Dismissed

`AGENTS.md` (Journal section) states:

> For individual journal fields (`enabled`, `snapshotInterval`,
> `contractIDs`, `redactions`, `diff`, `applyPatch`), `null` is rejected
> — only the documented value types are accepted; pass `undefined` (or
> omit the field) to leave it alone.

The code in `src/chelonia.ts:455-498` does not reject `null`. Each
branch only checks `!== undefined`, so a literal `null` is treated as a
real value and has confusing, field-specific effects:

| Field | Behaviour on `null` |
|---|---|
| `enabled: null` | `target.enabled = null` → `cfg?.enabled === true` is `false`, silently disables |
| `snapshotInterval: null` | falls back to default, warns `invalid snapshotInterval null` |
| `contractIDs: null` | `null?.slice()` → `target.contractIDs = undefined`, silently re-broadens to "all contracts" |
| `redactions: null` | `null?.map(...)` → `target.redactions = undefined`, **and** `sbp('chelonia/journal/clear')` runs anyway, wiping all journals |
| `diff: null` | `target.diff = null` (the read site's `?? defaultDiff` then masks it, but persisted config now lies) |
| `applyPatch: null` | same as `diff` |

The most surprising one is `redactions: null` — it triggers a full
journal wipe even though the caller didn't actually change the
redaction set.

Fix: either honour the documented contract and reject `null` per field
with a thrown `TypeError`, or update `AGENTS.md` to describe the actual
field-by-field behaviour. Since `null` is plainly invalid input,
rejecting is safer:

```ts
const rejectNull = (name: string, v: unknown) => {
  if (v === null) {
    throw new TypeError(
      `[chelonia][journal] config.journal.${name} cannot be null; ` +
      `omit the field to leave it alone`
    )
  }
}
rejectNull('enabled', journalOverride.enabled)
rejectNull('snapshotInterval', journalOverride.snapshotInterval)
// ...etc
```

---

## 2. 🔴 `journal: null` documented as no-op, but only works by accident

- [x] Addressed
- [ ] Dismissed

`AGENTS.md`:

> The top-level `config.journal` block itself can be omitted; passing
> `null` for the whole block is also treated as "leave alone".

In `src/chelonia.ts:445-453`:

```ts
const journalOverride = config.journal
const configForMerge = journalOverride !== undefined
  ? { ...config, journal: undefined }
  : config
merge(this.config, configForMerge)
// ...
if (journalOverride) { /* field-by-field overrides */ }
```

`journal: null` enters the `!== undefined` branch (journal stripped
out of `configForMerge`), then the override block is skipped because
`if (journalOverride)` is falsy. The end result is "leave alone", but
only because of two opposing guards lining up. Any future tweak that
changes the second guard to `!== undefined` (matching the first) turns
this into a `TypeError: Cannot read properties of null` on
`journalOverride.enabled`.

Recommend explicit normalisation at the top:

```ts
const journalOverride = config.journal == null ? undefined : config.journal
```

…and a dedicated test case pinning `journal: null` to "no-op".

---

## 3. 🟡 `redactions` provided with the same value still wipes all journals

- [x] Addressed
- [ ] Dismissed

`src/chelonia.ts:481-497`:

```ts
if (journalOverride.redactions !== undefined) {
  target.redactions = journalOverride.redactions?.map(r => ({
    path: r.path,
    redact: r.redact
  }))
  // ...
  sbp('chelonia/journal/clear')
}
```

The clear fires *unconditionally* whenever the caller passes
`redactions`, even when the new value is structurally identical to the
old one. In practice this means common patterns like "re-configure
with the full journal block to toggle `enabled`" silently destroy
journal history because the same `redactions` array is re-supplied.
The integration tests reset config in `beforeEach` by re-passing the
full block, so this footgun is invisible there.

Suggested fixes:

1. Structural equality check (cheap — `JSON.stringify` of `path` arrays
   plus reference-equality on `redact` functions) and only clear when
   the set actually changes.
2. Or gate the destructive clear behind an explicit opt-in flag and
   make callers responsible otherwise.

---

## 4. 🟡 `snapshotInterval` validation lives only in `configure`, not in the read path

- [ ] Addressed
- [x] Dismissed

`chelonia/configure` (`src/chelonia.ts:461-478`) carefully rejects
non-integer / non-positive / non-finite `snapshotInterval` values. But
the seeded default in `chelonia/_init` and the runtime read in
`resolveJournalConfig` (`src/journal.ts:535-542`) only check
`typeof === 'number'`:

```ts
const snapshotInterval = typeof cfg?.snapshotInterval === 'number'
  ? cfg.snapshotInterval
  : DEFAULT_SNAPSHOT_INTERVAL
```

So `Infinity`, `NaN`, `0`, `-1`, and fractions pass through if the
journal config is ever assembled outside `chelonia/configure` (caller
mutating `this.config.journal.snapshotInterval` directly, future
surface bypassing `configure`, etc.). Concretely:

- `Infinity`: `entries.length > 2 * Infinity` always false → never
  trims, journal grows unboundedly.
- `0`: snapshot pushed on every patch *and* `length > 0` always true
  → trim runs every event, can collapse the window.
- `NaN`: all comparisons false → never trims and never snapshots.

Move validation into a shared `normalizeJournalConfig` helper consumed
by both `configure` and `resolveJournalConfig`.

---

## 5. 🟡 `chelonia/journal/reconstruct` swallows applier failures and returns `undefined`

- [x] Addressed
- [ ] Dismissed

`src/journal.ts:792-798`:

```ts
try {
  state = cfg.applyPatch(state, e.patch)
} catch (err) {
  logJournalError(`reconstruct failed at entry ${i}`, err)
  return undefined
}
```

`reconstruct` is documented as "useful as a self-check". A self-check
that returns `undefined` (the same value used for "no journal exists")
on corruption is hard to act on — callers cannot distinguish "no
journal" from "journal is broken". For a debugging tool, you want a
loud failure mode.

Suggest one of:

- Throw a typed error (e.g. `ChelErrorJournalCorrupt`) instead of
  returning `undefined`, OR
- Return a discriminated union
  `{ ok: true, state } | { ok: false, reason, entryIndex }`.

---

## 6. ⚪ `appendAndTrim` mutates `existing` in place — observers see transient intermediate states

- [x] Addressed
- [ ] Dismissed

`src/journal.ts:567-573`:

```ts
// We mutate `entries` in place. `recordEvent` always replaces the
// wrapper object via `reactiveSet(contractMeta, '_journal', { entries })`
// so subscribers go through the wrapper and the array's identity is an
// implementation detail.
entries.push(entry)
```

`chelonia/journal/get` deep-clones, so external readers are safe. But
the in-place mutation has two subtler implications:

- Reactive frameworks observing the array directly (Vue 2/3 reactivity
  on `_journal.entries.length`) will fire on every `push` and on the
  `splice` trim — not on a single atomic swap. A consumer can observe
  the array at length 2X+1 *before* the trim splice runs.
- The comment says "subscribers go through the wrapper". That's only
  true if consumers dereference `_journal` on every read; anyone who
  does `const { entries } = _journal` holds the live array.

Either allocate a fresh array per call (simple, and matches the
reactive-set semantics for the wrapper), or document explicitly: "do
not retain `_journal.entries` across event ticks".

---

## 7. ⚪ `_journal` persisted under `state.contracts[id]` — can dominate per-contract storage

- [ ] Addressed
- [x] Dismissed

`AGENTS.md` flags this, but with default `snapshotInterval = 50` the
journal carries up to ~50 diffs plus two redacted full-state snapshots
per active contract — which for many real contracts is larger than the
contract state itself. For an opt-in debug feature that is acceptable,
but:

- There is no first-event warning when enabling on a long-lived
  contract.
- Once enabled, retroactively trimming requires `chelonia/journal/clear`
  (which loses history).

Consider:

- A `chelonia/journal/stats` selector (count + approximate JSON byte
  size) so consumers can monitor without manually walking `_journal`.
- A documented "for production-leaning deployments, set
  `snapshotInterval ≤ 10`" guideline.

---

## 8. ⚪ `description` is persisted verbatim and unredactable except by post-hoc scrubbing

- [ ] Addressed
- [x] Dismissed

`src/journal.ts:643-645`:

```ts
let description: string | undefined
try { description = message.description() } catch { /* optional */ }
```

`AGENTS.md` correctly warns this can leak action data for unencrypted
ops. A defence-in-depth improvement: allow the journal config to
provide a `describe?: (m: SPMessage) => string | undefined` override
(defaulting to `m.description()`), so consumers worried about leakage
can plug in a sanitiser at record time without resorting to
post-`get` scrubbing of persisted state.

---

## 9. ⚪ `shortHashRedactor` is documented as unsafe for low entropy, but bundled and exported

- [ ] Addressed
- [x] Dismissed

The redactor is exported from the package root (and the `./journal`
subpath) and is explicitly called out in `AGENTS.md` as trivially
reversible for low-entropy values. The *typical* contract-state keys
consumers want to redact (`type`, status enums, booleans like
`isAdmin`) are exactly the low-entropy cases the comment warns
against, so the named export is a footgun.

At minimum, rename to signal the limitation —
`shortHashRedactor_unsafeForLowEntropy` — or ship a paired
`constantSentinelRedactor` so the "right answer for booleans" is
equally discoverable.

---

## 10. ⚪ `cloneValue` aliases `Date` / `Map` / `Set` / class instances across journal entries

- [ ] Addressed
- [x] Dismissed

`src/journal.ts:69-95` falls through on non-plain values and returns
them by reference. The doc mentions this caveat, but the failure mode
in practice — silent, retroactive mutation of historical entries the
moment any contract handler in-place-mutates a `Date`/`Map`/`Set` on
`state` — is severe for a feature whose whole job is to be a
trustworthy debugging record.

Either:

- Treat non-plain values as opaque and JSON-serialise them at clone
  time (lossy but safe), OR
- Detect them and emit a one-time `console.warn` per `recordEvent`
  call so consumers notice before deploying.

---

## 11. ⚪ `chelonia/config` snapshot returns the live `redact` functions by reference

- [ ] Addressed
- [x] Dismissed

`src/chelonia.ts:415-426`:

```ts
out.journal = {
  ...out.journal,
  redactions: this.config.journal.redactions?.map(r => ({
    path: r.path,
    redact: r.redact
  })),
  diff: this.config.journal.diff,
  applyPatch: this.config.journal.applyPatch
}
```

The comment acknowledges `redact` is shared by reference. That's the
right call for *functions*, but combined with the round-trip pattern
(`get config → mutate → configure`) it means a caller who treats the
snapshot as a deep copy and reassigns `redact` will mutate the live
function reference shared with anyone else holding the snapshot. Not a
bug per se — but worth a one-line "snapshot is field-shallow on
functions" note in the comment so the contract is explicit at the
callsite.

</details>

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/26185707589)._

src/chelonia.ts
```diff
@@ -420,18 +437,207 @@ export default sbp('sbp/selectors/register', {
...
+        })),
+        diff: this.config.journal.diff,
+        applyPatch: this.config.journal.applyPatch
+      }
     }
+    return out
   },
   'chelonia/configure': async function (this: CheloniaContext, config: CheloniaConfig) {
-    merge(this.config, config)
+    // Pull the journal override aside before `merge`: `merge` deep-clones
```
taoeffect, line 469: Could you please add documentation on how to use the journal in an app?

Either here, or in README.md, or perhaps in both locations? It should show `'chelonia/configure'` being called with the parameters, and show which fields are required and which are optional

corrideat: Hm, I could, but this is already defined in `types.ts`:

```
export type JournalConfig = {
  enabled?: boolean;
  snapshotInterval?: number;
  // When omitted or empty, applies to all contracts (provided `enabled` is
  // true). Otherwise only listed contractIDs are journaled.
  contractIDs?: string[];
  redactions?: JournalRedaction[];
  diff?: (before: unknown, after: unknown) => JournalPatch[];
  applyPatch?: (state: unknown, patches: JournalPatch[]) => unknown;
};
```

I'm not against adding the documentation here, but just wondering about the _how_, since we already have lots of options for `chelonia/configure`, none of which are documented, such as `connectionURL`, `stateSelector`, `fetch`, `skipActionProcessing`, just to name a few.

It seems to me like we could add documentation to the `README.md` file, but that we should document all or most fields, not just the journal (which is less important for a functioning app than, e.g., `connectionURL`).

taoeffect: @corrideat might be a good opportunity to create a `docs/` folder and add it there. This btw is something an LLM can do really well.

corrideat: Ok, so, are you saying we add a `docs` directory documenting `chelonia/configure` in general (including the journal), and we reference that in the `README.md` file?

taoeffect: Yes, I think that would be a good idea, and should be equivalent effort to just documenting the journal options thanks to the LLM doing most/all of the work.

src/chelonia.ts
```diff
@@ -420,18 +437,207 @@ export default sbp('sbp/selectors/register', {
...
+        // `{ path, redact }` objects with the live config, so a caller
+        // could mutate the live redaction path via the returned snapshot.
+        // The `redact` function is intentionally shared by reference
+        // (functions can't be cloned).
+        redactions: this.config.journal.redactions?.map(r => ({
+          path: r.path,
+          redact: r.redact
+        })),
+        diff: this.config.journal.diff,
+        applyPatch: this.config.journal.applyPatch
```
taoeffect, line 463: Is this how the journal is retrieved?

If so, I'm not a huge fan of this. Could we please have a dedicated selector for retrieving it?

The the config selector is for retrieving the config, not the config plus other internal stuff. Having the journal itself be retrieved would be like asking the config to give us the chelonia state - makes no sense.

corrideat: I agree, but this is about the journal config, not about retrieving the journal. The journal itself is retrieved using 'chelonia/journal/get'

taoeffect: I see, sorry I missed that!

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/26296668236)._

taoeffect:
GPT-5.5:

<details>

## 1. 🔴 Package version is downgraded from `main`

- [ ] Addressed
- [x] Dismissed

`package.json:3` and `package-lock.json:3` set the package version to `1.4.3`, while `main` is already at `1.4.4`. Merging this branch as-is would move the package metadata backwards, which can break release automation and make the journal feature unpublishable or indistinguishable from an older release.

Restore the version from `main` or bump forward from it, and keep `package-lock.json` in sync:

```json
{
  "name": "@chelonia/lib",
  "version": "1.4.4"
}
```

## 2. 🟡 Public `chelonia/in/deserializeMessage` selector was removed

- [ ] Addressed
- [x] Dismissed

`src/chelonia.ts:1654` now jumps directly from `chelonia/contract/disconnect` to `chelonia/in/processMessage`; the existing public selector `chelonia/in/deserializeMessage` that lived immediately before `processMessage` on `main` was deleted. That selector was documented in code as a debugging/low-level primitive for applications that need to deserialize raw `SPMessage`s, so removing it is a breaking API change unrelated to the journal feature.

Re-add the selector unless this branch intentionally makes a breaking API removal:

```ts
'chelonia/in/deserializeMessage': function (
  this: CheloniaContext,
  rawMessage: string,
  options: { headOnly?: boolean, state?: ChelContractState } = {}
) {
  if (options.headOnly) {
    return SPMessage.deserializeHEAD(rawMessage)
  } else {
    return SPMessage.deserialize(
      rawMessage,
      this.transientSecretKeys,
      options.state,
      this.config.unwrapMaybeEncryptedData
    )
  }
},
```

</details>

taoeffect (approved):
Great work!!

src/chelonia.ts
```diff
@@ -420,18 +437,207 @@ export default sbp('sbp/selectors/register', {
...
+      // null/empty value to mean "off".
+      this.config.journal = {
+        enabled: false,
+        snapshotInterval: DEFAULT_SNAPSHOT_INTERVAL,
+        contractIDs: [],
+        redactions: []
+      }
+      sbp('chelonia/journal/clear')
+    } else if (journalOverride !== undefined) {
+      // Reject `null` per individual field: AGENTS.md documents that
```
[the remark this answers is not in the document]
corrideat: dismissed

src/types.ts
```diff
@@ -89,6 +89,81 @@ export type CheloniaConfig = {
...
+      // The event's `SPMessage.description()` output (raw, never passed
+      // through `redactions`). For unencrypted ops this can include the
+      // action name and action-data fragments; treat it as journal-visible.
+      // Callers worried about leakage should either strip `description`
+      // before persisting `entries` or rely exclusively on encrypted ops.
+      description?: string;
+      // Redacted deep clone of the per-contract state AFTER this event was
+      // processed. May be `null` if the contract state was undefined (e.g.,
+      // failed first-message processing).
+      state: unknown;
```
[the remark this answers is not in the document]
corrideat: dismissed

PR #79, opened 2026-05-26
Improve re-ingest logic

### LINKED ISSUE #77: Bad previousHEAD errors when resyncing
### Problem

Upon re-syncing (manually through the 'Re-sync and rebuild data' troubleshooting tool), a user saw this error:

<figure>

Image

<figcaption>ChelErrorUnrecoverable('more than 100 different bad previousHEAD errors')</figcaption>

</figure>

Things reportedly work afterwards, so it looks like a transient issue.

### Solution

Identify the cause of this issue and fix it.

corrideat:
Closes #77

### Potentially breaking changes

Before, there was a global list of events to re-ingest of 100 events (shared among all contracts). Now, this list is per-contract and capped at 20 events. This seems reasonable, but it also means we might throw in situations previously we didn't and vice-versa.

### AI disclosure

Assisted by Opus 4.7


taoeffect:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/26476961907)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/26513487376)._

corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/26577150509)._

src/internals.ts
```diff
@@ -525,7 +530,7 @@ export default sbp('sbp/selectors/register', {
       // pass in globals that we want access to by default in the sandbox
       // note: you can undefine these by setting them to undefined in exposedGlobals
       crypto: {
-        getRandomValues: <T extends ArrayBufferView>(v: T) =>
+        getRandomValues: <T extends ArrayBufferView<ArrayBuffer>>(v: T) =>
```
corrideat, line 534: Unrelated to the issue in this PR, but it has to do with changes to TypeScript built-in types.

src/internals.ts
```diff
@@ -3088,14 +3104,63 @@ export default sbp('sbp/selectors/register', {
...
+    // ever changes (e.g. a single debounce shared across contracts),
+    // the capture must be made explicit (e.g. via `.bind`) to avoid
+    // firing a force-sync against the wrong contract.
+    d = debounce(() => {
+      // Drop our slot before firing so a follow-up call after the timer
+      // has fired but before `chelonia/private/out/sync` resolves gets a
+      // fresh debounce rather than reusing a stale one (whose internal
+      // `timeout` is already `undefined` so it'd fire instantly).
+      reprocessDebounceMap.delete(contractID)
+      sbp('chelonia/private/out/sync', contractID, { force: true }).catch((e: unknown) => {
```
taoeffect: I'm noticing that we have a selector here called `'chelonia/private/out/sync'`. And indeed that selector was there before.

But this seems incorrectly named. Why is it "out"? We aren't sending anything out, and indeed that selector just wraps various "in" operations like `'chelonia/private/in/syncContract'`.

So, perhaps we should take this opportunity to rename this to `'chelonia/private/in/sync'` and move it lower down in `internals.ts` next to the other `in` selectors:

Image

corrideat: It's sending something 'out', i.e., making external requests as opposed to taking something 'in' that came, e.g., through the WebSocket. However, this is inconsistently applied, and the only 'out' part of this selector is calling an 'in' selector.

What's confusing is that we don't have a proper definition of 'out' or 'in' (and how to resolve things that are both 'out' and 'in')

Since we're not sending an event out, which seems to be the main convention for the out-in naming, I agree that the name is wrong.

taoeffect: I see. Well, I do think it needs to be renamed, but perhaps as an alternative we could just call it `'chelonia/private/sync'`.

taoeffect:
Great work! Note the `dist/` build is outdated so I had to rebuild it locally (for whatever reason Opus noticed it... it wasn't supposed to because `code-review` tells it to ignore that folder...).

Opus 4.8 Review

<details>

## 1. 🟡 Per-contract `reprocessDebounced` rationale comment is factually wrong (and the same path drops debounce coalescing semantics)

- [x] Addressed
- [ ] Dismissed

`src/internals.ts:3133-3142`. The new per-contract debounce deletes its map slot
*before* the wrapped function fires, justified by this comment:

```ts
// Drop our slot before firing so a follow-up call after the timer
// has fired but before `chelonia/private/in/sync` resolves gets a
// fresh debounce rather than reusing a stale one (whose internal
// `timeout` is already `undefined` so it'd fire instantly).
reprocessDebounceMap.delete(contractID)
```

The parenthetical is incorrect. Re-reading turtledash's `debounce`
(`node_modules/turtledash/dist/esm/index.js:232-245`): when `later()` fires it sets
`timeout = undefined`, and the next call to `debounced()` hits `if (!timeout) timeout
= setTimeout(later, wait)` — i.e. it **re-arms with a fresh full `wait`**, it does
*not* "fire instantly". (Firing instantly only happens with `immediate: true`, which
is not used here.) So a reused "stale" debounce would behave correctly; the delete is
harmless but the stated reason is wrong, and a future reader may make decisions based
on it.

There is also a real, if minor, behavioral consequence of the delete-before-fire that
is worth calling out explicitly: it changes the de-dup window. With the old single
`debounce(fn, 1000)`, repeated `reprocessDebounced(cid)` calls coalesced into one
force-sync per 1 s *rolling* window indefinitely. Now, as soon as the 1 s timer
fires, the slot is removed; the *next* gap event for that contract immediately
creates a brand-new debounce and schedules another force-sync 1 s later. That's
arguably fine (each fired sync is followed by at most one more), but it's a different
contract than "coalesce while busy", and the wrong comment hides that.

Suggestion: fix the comment to state the actual reason (avoid an unbounded-lifetime
entry lingering in the map / make each fired cycle independent), e.g.:

```ts
// Remove our map slot before firing. The debounce has already fired
// (its internal timeout is now cleared), so keeping it would just leave
// a dead entry in the map until the next gap event re-arms it. Dropping
// it now means each force-sync cycle is self-contained and the map only
// holds contracts with a currently-pending timer.
reprocessDebounceMap.delete(contractID)
```

---

## 2. ⚪️ `reprocessDebounced` is not cancelled on successful re-ingest, so a redundant force-sync can still fire

- [ ] Addressed
- [x] Dismissed

`src/internals.ts:3204-3219` (`checkMessageOrdering`). The success path drops the
tracker entry via `noteReingestSuccess`, but does **not** cancel a still-pending
`reprocessDebounced` timer for that contract:

```ts
if (noteReingestSuccess(contractID, hash)) {
  console.warn(`[chelonia] WARN: successfully reingested ${message.description()}`)
}
```

Sequence: a gap event schedules a force-sync (debounce armed, ~1 s). If the missing
events arrive over the live subscription before the timer fires and the gap event is
re-delivered and processed successfully, the tracker is cleaned but the debounce
timer still fires a `chelonia/private/in/sync(cid, { force: true })`. That's a wasted
network round-trip, not a correctness bug (force-sync of an up-to-date contract is a
no-op), so this is low severity. If you want to tighten it, call
`clearReprocessDebounceForContract(contractID)` in the success branch.

---


## 3. ⚪️ Forward-declared `const` helpers used before their definition rely on call-time-only invocation

- [ ] Addressed
- [x] Dismissed

`clearReprocessDebounceForContract` is invoked at `src/internals.ts:629` (inside the
`chelonia/private/removeImmediately` selector) but defined as a module-level `const`
arrow at `src/internals.ts:3152`. This is safe **only** because `removeImmediately` is
a runtime-invoked SBP selector — by the time it runs, the module has finished
evaluating and the `const` is initialized, so there's no temporal-dead-zone hit.

This is correct as written, but it's a latent footgun: if anyone ever calls
`clearReprocessDebounceForContract` (or `reprocessDebounced`) from module top-level
init code above line 3152, it will throw a TDZ `ReferenceError`. A short comment at
the call site, or hoisting these helpers above the `sbp('sbp/selectors/register', …)`
block, would remove the surprise. Low priority.

---

## 4. ⚪️ `noteRecingestSuccess` only fires on the exact-next-height path — a gap that is closed by a re-sync replay never clears its tracker entry, slowly consuming the per-contract cap

- [x] Addressed
- [ ] Dismissed

In `handleEvent.checkMessageOrdering` (`src/internals.ts:3229`):

```ts
if (noteReingestSuccess(contractID, hash)) {
  console.warn(`[chelonia] WARN: successfully reingested ${message.description()}`)
}
```

This runs only for the in-order case (`height === latestProcessedHeight + 1`,
i.e. after the two early branches for already-processed and future events). When
a forced re-sync backfills a gap, the messages are typically replayed in order
through `handleEvent`, so the recorded future-hash *is* eventually re-delivered
at the right height and cleared here — that is the intended happy path and it
works.

However, after a `removeImmediately(cid, { resync: true })` the contract is
re-fetched from height 0 and the **previously-recorded future hash may never be
re-delivered to `checkMessageOrdering` at the success height** (e.g. the message
was a transient fork/orphan that no longer appears on the canonical chain, or it
arrives via a path that short-circuits before the success branch). The PR
already handles the resync case by clearing the whole per-contract tracker in
`removeImmediately` (`src/internals.ts:628`), which is the right safety net — so
this is mostly defense-in-depth.

The residual concern: in steady state (no reset/resync), if a recorded gap hash
is *never* the one that closes the gap (the gap is filled by a different
message), that hash lingers in the per-contract `Set` forever, monotonically
consuming the `REINGEST_PER_CONTRACT_CAP = 20` budget until the contract trips
`ChelErrorUnrecoverable` for what is effectively accumulated dead state rather
than a genuine wedge. The old global code had the same shape of leak, so this
isn't a regression — but since the cap dropped from 100 (global) to 20
(per-contract), the margin before a long-lived, occasionally-gappy contract hits
the cap is much smaller now.

Consider one of:

- Clearing the per-contract tracker whenever the contract successfully advances
  past the recorded height (not just when the exact recorded hash is the one
  that advances it), e.g. prune entries whose recorded height is now
  `<= latestProcessedHeight` inside the success branch; or
- Documenting that `REINGEST_PER_CONTRACT_CAP` is a "distinct unrecovered gaps
  over the contract's *session* lifetime" budget, not "concurrent pending gaps",
  so the 20 value is chosen with that in mind.

Note the tracker keys on `hash` only, while the cap reasoning is about
`height`; adding the height to the stored value would make height-based pruning
straightforward.

</details>

corrideat:
>  build is outdated so I had to rebuild it locally

I don't believe it was _outdated_, but some files were missing. It should now be fixed.

taoeffect (approved):
Great work @corrideat! 👏 👏  (And Opus!)

PR #80, opened 2026-05-27
26 enhanced kv store key value experience

### LINKED ISSUE #26: Enhanced KV-store / Key-value experience
### Problem

Creating apps that use the KV store is a bit too cumbersome.

1. It requires having pubsub listeners in different files
2. It requires complex handlers that take `onConflict` parameters and `etag` stuff, e.g. `const getUpdatedUnreadMessages = ({ currentData = {}, etag } = {}) => {`
3. We're currently adding multiple selectors for every type of KV we are manipulating

### Solution

We want a very simple API that is super easy to use for handling KV updates that uses Chelonia to automatically decide where in the state these  keys are stored, and creates a simple API for creating updates and handling conflicts.

See https://github.com/okTurtles/group-income/issues/2903 for details.

Also see the convenience function `updateKVPreferences` from Sebin's recently merged PR in `identity-kv.js`.

The user should not be responsible for anything related to pubsub events when dealing with KV stuff.

The pubsub KV logic could still be available for more complicated use-cases.

EDIT: since we would be moving most of the contract actions into the KV store (e.g. `state.attributes`) we should also think about how this new friendlier KV API would handle validation. So we could for example say that "whenever this key is validated please use this [zod](https://zod.dev/) schema to validate".

corrideat:
Closes #26 

~~[KV-REVAMPED.md](https://github.com/user-attachments/files/28322374/KV-REVAMPED.md)~~ ~~[KV-REVAMPED.md](https://github.com/user-attachments/files/28480257/KV-REVAMPED.md)~~ 
~~[KV-REVAMPED.md](https://github.com/user-attachments/files/29213010/KV-REVAMPED.md)~~ 
[KV-REVAMPED.md](https://github.com/user-attachments/files/30245870/KV-REVAMPED.md) contains the spec used (which this PR should conform to).

### AI disclosure

GPT-5.5, Opus 4.7 and GLM 5.1 were used in generating code for this PR. GPT-5.5, Opus 4.7 & GLM 5.2 were used for reviews.


corrideat:
/crush_fast

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/26536512912)._

src/chelonia.ts
```diff
@@ -2573,21 +2667,31 @@ export default sbp('sbp/selectors/register', {
...
   ) {
     maxAttempts = maxAttempts ?? 3
     const url = `${this.config.connectionURL}/kv/${encodeURIComponent(contractID)}/${encodeURIComponent(key)}`
     const hasOnconflict = typeof onconflict === 'function'
+    // Compose the caller-provided signal with Chelonia's global abort
+    // controller (KV-REVAMPED.md §4.2). When `callerSignal` is omitted,
+    // we fall back to the global signal unchanged for full backwards
+    // compatibility with existing call sites.
+    const fetchSignal: AbortSignal = callerSignal
+      ? AbortSignal.any([this.abortController.signal, callerSignal])
```
corrideat: This (`AbortSignal.any`) is a relatively new API with somewhat limited support (all modern browsers support it, but as of today it has [89.1%](https://caniuse.com/mdn-api_abortsignal_any_static) global support.)

Since this is new functionality and therefore not something we are currently using, and something that we may not need to use, perhaps it's fine.

However, we should consider rewriting this or using a shim in GI.

src/kv.ts
```diff
@@ -0,0 +1,2453 @@
...
+    ) {
+      const sorted: Record<string, unknown> = {}
+      for (const k of Object.keys(val).sort()) {
+        sorted[k] = (val as Record<string, unknown>)[k]
+      }
+      return sorted
+    }
+    return val
+  })
+}
```
taoeffect: Pretty sure we already have this in turtledash as `hashableRepresentation`.

And if `hashableRepresentation` isn't sufficient then this needs to be moved into turtledash (and DRY'd with respect to hashableRepresentation)

src/kv.ts
```diff
@@ -0,0 +1,2453 @@
...
+      const status = perContract?.[slotKey]?.status ?? 'non-init'
+      if (status === 'error') return 'error'
+      if (status === 'loading') sawLoading = true
+      else if (status === 'non-init') sawNonInit = true
+      else if (status === 'loaded') sawLoaded = true
+    }
+    if (sawLoading) return 'loading'
+    if (sawNonInit) return 'non-init'
+    if (sawLoaded) return 'loaded'
+    return 'non-init'
```
taoeffect: Seeing a lot of use of string literals. Those should be `const` variables.

src/kv.ts
```diff
@@ -0,0 +1,2453 @@
...
+    }
+    const types = Array.isArray(def.contractType) ? def.contractType : [def.contractType]
+    if (types.length === 0) {
+      throw new ChelErrorKvSlotInvalid('[chelonia/kv] defineSlot: contractType required')
+    }
+    // Runtime validation of optional fields (SBP selectors are callable
+    // from JavaScript without TypeScript enforcement).
+    if (def.match != null && typeof def.match !== 'function') {
+      throw new ChelErrorKvSlotInvalid('[chelonia/kv] defineSlot: match must be a function')
+    }
```
corrideat, line 1191: Consider using Zod for schema validation.

corrideat:
Testing this, I noticed this for `notifications.spec.js` (which also seems to be transiently failing):

```
[19:46:48.781] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 204
[19:46:48.828] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 412
[19:46:49.521] DEBUG (62571): [sbp] backend/server/updateSize ["zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9",128]
[19:46:49.523] DEBUG (62571): [pubsub] Broadcasting KV change on zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9 to key notifications
[19:46:49.523] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 204
[19:46:49.576] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 412
[19:46:50.161] DEBUG (62571): [sbp] backend/server/updateSize ["zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9",128]
[19:46:50.165] DEBUG (62571): [pubsub] Broadcasting KV change on zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9 to key notifications
[19:46:50.165] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 204
[19:46:50.207] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 412
[19:46:51.005] DEBUG (62571): [sbp] backend/server/updateSize ["zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9",124]
[19:46:51.007] DEBUG (62571): [pubsub] Broadcasting KV change on zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9 to key notifications
[19:46:51.008] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 204
[19:46:51.048] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 412
[19:46:52.006] DEBUG (62571): [sbp] backend/server/updateSize ["zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9",128]
[19:46:52.007] DEBUG (62571): [pubsub] Broadcasting KV change on zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9 to key notifications
[19:46:52.008] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 204
[19:46:52.048] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 412
[19:46:52.691] DEBUG (62571): [sbp] backend/server/updateSize ["zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9",128]
[19:46:52.695] DEBUG (62571): [pubsub] Broadcasting KV change on zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9 to key notifications
[19:46:52.695] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 204
[19:46:52.763] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 412
[19:46:52.789] DEBUG (62571): [pubsub] Pinging clients
[19:46:53.524] DEBUG (62571): [sbp] backend/server/updateSize ["zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9",124]
[19:46:53.528] DEBUG (62571): [pubsub] Broadcasting KV change on zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9 to key notifications
[19:46:53.528] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 204
[19:46:53.570] DEBUG (62571): 127.0.0.1: POST /kv/zLDXeQ2AgfCuPcNQvBuix2oY8Zc1xH3nKTLeQShRVN1k8wACNT1RB3n9/notifications --> 412
```

Increasing the FIFO queue size from 8 to 128 seems to fix the failing tests, but we shouldn't be having so many conflicts writing to the notifications queue, and this points to a likely bug in the echo-suppression logic this PR implements. If we indeed have buggy logic, it needs to be fixed (to determine this, we need to examine actual update calls to see what's being written). The logs suggest that the client is confused about what should be written / updated and attempts to repeatedly make the same update.

This PR introduces a nonce in KV writes (which are then put in a FIFO queue) so that when we receive a value over pubsub, we can tell whether it's the value we've just written or an update made by someone else. However, we likely don't need this additional nonce (which is stored in the KV value) since we already hash KV values and use the `etag` / `x-cid` headers to expose these hashes to clients. However, we don't send this `x-cid` / `etag` over pubsub. Two possible solutions:

1. Send this information over pubsub (requires changing `chel serve`)
2. (less preferred) Have clients compute the `x-cid` value. This is less preferred because it requires the clients and the server agree on how the `x-cid` value should be computed (exact algorithm).


<ins>This seems resolved now, an it was caused by `POST /kv/...` not returning the ETag value, which is now in https://github.com/okTurtles/chel/pull/140</ins>

src/chelonia.ts
```diff
@@ -429,12 +431,30 @@ export default sbp('sbp/selectors/register', {
...
+    // because slot definitions are application code. The other four
+    // are per-subscription runtime state and are reset alongside
+    // `subscriptionSet` / `rootState._kv`.
+    this.kvSlots = new Map()
+    this.kvSlotsByContractID = new Map()
+    this.kvActiveFilters = new Map()
+    this.kvFilterDirty = new Set()
+    this.kvLocalEchoCIDs = new Map()
+    this.kvPendingWrites = new Map()
+    this.defContractKvByManifest = new Map()
```
taoeffect, line 453: All of the other variables in this section begin with `kv`, kinda indicating that they're together as part of the same system, except for this last one, which also has `Kv` in the name but it's not at the start of the variable name. For consistency, would it make sense to move that to the start of the variable name, e.g. `this.kvDefConractByManifest`, or is it intended to be set apart from the others?

src/chelonia.ts
```diff
@@ -1146,6 +1164,63 @@ export default sbp('sbp/selectors/register', {
...
+                    msg: [string, ParsedEncryptedOrUnencryptedMessage<object>],
+                  ) => void
+                ).call(this.pubsub, [msg.key, parsed])
+              } catch (e) {
+                console.error(
+                  `[chelonia] legacy kv pubsub callback threw for ${msg.channelID}::${msg.key}`,
+                  e
+                )
+              }
+            }
```
taoeffect: Don't quite understand what this "legacy layer" thing is. Do we need it or can it be removed?

corrideat: Whether we 'need' it or not depends on what the API should look like. This is used for supporting the existing way of handling KV updates in Group Income which doesn't use `chelonia/kv/update`.

If we were to remove it, then application developers won't have easy access to these lower level primitives to handle KV updates directly.

src/chelonia.ts
corrideat: It's different from the others in the sense that it's only used in / by `'chelonia/defineContract'`, which mostly follows that specific naming convention (`this.defContractSelectors`, `this.defContract`).

OTOH, you're right that we do have all of these other keys starting with `kv` when they are KV-related.

src/chelonia.ts
taoeffect: Letting developers have access to low-level APIs and doing backwards compatibility when it's not needed are two different things.

The word "legacy" sounds like backwards compatibility. If it's not backwards compat then it should probably be renamed. If it is backwards compat then it should be removed.

taoeffect:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/28888414796)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/29102080555)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/29114462953)._

taoeffect:
Fable review ready!

<details>

# Code Review

**Base**: `main`
**Head**: `pr` (da28cd8)
**Date**: 2026-07-10
**Model**: Fable (Claude Fable 5)
**Scope**: KV API revamp (KV-REVAMPED.md) — `src/kv.ts` (new, 3496 lines), `src/kv.test.ts` (new, 156 tests), plus extensions to `chelonia.ts`, `chelonia-utils.ts`, `types.ts`, `local-selectors/`, `pubsub/`, `events.ts`, `errors.ts`, `internals.ts`, docs.

---

## Verification performed

- Read the full KV-REVAMPED.md design document and checked the implementation against it (§4 API surface, §11.2 index invariant, §11.3 selector semantics, §11.5 filter coalescing, §4.9 self-echo suppression).
- `npm test` passes (kv.test.ts suite + 133 aggregate tests), `npm run lint` clean.
- `npm run build` re-run from scratch: **zero drift** against the committed `dist/` output (the `Build` commit is complete; the subsequent merge of `main` brought no `src/` changes).
- Two targeted call-site investigations: (a) pubsub `kvFilter` restoration across connect/reconnect, (b) all four `CONTRACTS_MODIFIED` emission sites vs. the new KV listeners.

## Compatibility assessment (changes outside `kv.ts`)

Per request, breaking-change analysis was done first. **No breaking changes found.** Details of every observable change to pre-existing surface:

| Change | Verdict |
|---|---|
| `chelonia/kv/set` resolves to `{ etag: string \| null }` (was `void`); accepts `signal`, forwards `maxAttempts` | Additive. All existing callers ignore the return value. |
| `chelonia/kv/set` conflict exhaustion now throws `ChelErrorKvMaxAttempts` (internal class) instead of plain `Error` | Same message string (`'kv/set conflict setting KV value'`); only `.name`/prototype differ. Documented in `internal-errors.ts`. Callers matching on message are unaffected. |
| `chelonia/kv/set` recovery GET on body-less 409/412 (`chelonia.ts`, `recoveryGetAttempted`) | New behavior, strictly better for direct callers: `onconflict` now receives a recovered `etag`/`currentData` instead of `undefined`, bounded to one GET per call (locked in by `chelonia-kv-set.test.ts`). |
| Conflict with no `onconflict` handler now throws before the 0–1500 ms random delay instead of after | Same error, faster failure. Not observable except in timing. |
| `chelonia/kv/get` returns `ChelKvGetResult` (parsed message + lazily attached `etag`) | Additive; `defineProperty` on the (unfrozen) result object, laziness of the `data` accessor preserved. |
| `chelonia/kv/queuedSet` forwards `signal`/`maxAttempts`, returns `{ etag }` | Additive. |
| `ChelKvOnConflictCallback` return type widened to `[JSONType, string \| undefined] \| false` | Type-level only. The runtime `if (!result) return false` falsy-abort path already existed on `main`; the widened type documents pre-existing behavior (the "behaviour change" note in `types.ts` slightly overstates this — see issue 6). |
| `NOTIFICATION_TYPE.KV` message handler restructured (`chelonia.ts:1170+`) | Verified: a consumer-provided raw KV handler is still invoked with `[key, parsed]` inside the same per-contract `queueInvocation` lane; the slot layer runs *in addition*. When neither a raw handler nor the kv module exists, the handler returns before parsing. No regression. |
| `parseEncryptedOrUnencryptedMessage` throws `ChelErrorInvalidMessageHeight` (was plain `Error`) | Name change only; exported and documented in `docs/api.md`. |
| `chelonia/externalStateSetup` now uses `sbp(stateSelector)` in the `CONTRACTS_MODIFIED` handler (was hardcoded `state/vuex/state`) | Bug fix. Behavior changes only for consumers that passed a `stateSelector` different from `'state/vuex/state'` — previously their contract projections silently went to Vuex. Worth a line in release notes. |
| `chelonia/externalStateSetup` returns a teardown function | Additive (previously returned `undefined`). |
| `chelonia/reset`: abort moved before `postCleanupFn`, KV writes drained via `_waitInFlight` | In-flight fetches are now aborted earlier (before `postCleanupFn` instead of after). All contract waits still precede it; low risk. |
| `chelonia/contract/fullState` gains optional `key` param and `kvState`/`kvEntry` result fields | Additive. |
| `chelonia/defineContract` throws if `contract.kv` is present and the kv module isn't loaded | Only affects new-style contracts; correct fail-fast. |
| `createKvMessage` gains optional `cid`; `NOTIFICATION_TYPE.KV` msg type gains `cid?` | Additive. Old servers that emit frames without `cid` degrade gracefully (authoritative-GET path, documented in `docs/api.md`). |
| `index.ts` import reordering (`chelonia.ts` last, because it locks the `chelonia` SBP domain) | Consistent with the pre-existing constraint that already applied to `./journal`, `./files`, etc. Deep-importing `./chelonia` before `./kv` silently drops the kv selector registration (SBP warns) — pre-existing pattern, now explicitly documented in `index.test.ts`. |

Cross-cutting wiring verified specifically:

- **Filter restoration across connect/reconnect**: `chelonia/connect` pre-seeds `client.kvFilter` synchronously before the socket can open, and the pubsub client reads `kvFilter` lazily at SUB-send time (with a KV_FILTER re-send on SUB-ack mismatch), so filters survive first connect and every reconnect. The direct-map-mutation idiom at `chelonia.ts:1244` is only safe because the socket is guaranteed not-yet-open there (it does not transmit) — fine as used, but do not reuse that idiom against a connected client.
- **`CONTRACTS_MODIFIED`**: all four emit sites pass `Array.from(this.subscriptionSet)` (matches the new listener typing), mutate `subscriptionSet` before emitting (the design's ordering claim holds), and the direct `_cleanupContractRuntime` call in `removeImmediately` plus the listener-driven call are correctly idempotent.
- **Self-echo suppression / conflict-marker state machine** (`_handleRemote`): traced both arrival orderings (echo-first and competing-frame-first), the demote-don't-delete rule, TTL expiry, and the eviction cap; no mirror-regression path found. Suppression is recorded before the abort/staleness checks in `update`/`clear`, so a committed write's echo can never re-validate against a replaced slot.
- **Lane serialization**: mirror reads, reducers, `kv/set`, `_handleRemote`, and `onUpdate` all run inside the per-contract `queueInvocation` lane; `_waitInFlight`'s three-source drain (index, echo CIDs, pending-write counters) covers the queued-but-not-started window.

The issues found follow, in priority order.

---

## 1. 🟡 Filter flush enters an infinite 2-second retry loop if slots reconcile before `chelonia/connect`

- [x] Addressed
- [ ] Dismissed

`chelonia/kv/setFilter` dereferences `this.pubsub` unguarded (`src/chelonia.ts:3023-3025`):

```ts
'chelonia/kv/setFilter': function (this: CheloniaContext, contractID: string, filter?: string[]) {
  this.pubsub.setKvFilter(contractID, filter)
},
```

`this.pubsub` is only assigned inside `chelonia/connect` (`src/chelonia.ts:1086`). Before this PR that was harmless — only consumers called `setFilter`, and they did so after connecting. Now the slot layer calls it autonomously: `defineSlot` reconciles against already-synced contracts and queues a filter flush (`src/kv.ts:1398-1408` → `queueFilterFlush` → `flushDirtyFilters`). If an app syncs contracts (HTTP-only, works without a websocket) and registers slots before calling `chelonia/connect` — or never connects at all (SSR / CLI / tooling contexts) — the flush throws `TypeError: Cannot read properties of undefined (reading 'setKvFilter')`, which is caught at `src/kv.ts:918-928` and routed to `scheduleFilterRetry`:

```ts
} catch (e) {
  console.warn(`[chelonia/kv] setFilter flush failed for ${cID}`, e)
  ...
  scheduleFilterRetry(ctx, cID)
}
```

`scheduleFilterRetry` (`src/kv.ts:942-956`) re-queues the contract and re-flushes after `KV_FILTER_RETRY_MS` (2 s), fails identically, and re-enters itself — an unbounded retry loop that (a) spams one `console.warn` per contract every 2 seconds and (b) keeps a live `setTimeout` chain that prevents a Node process from ever exiting. Nothing distinguishes this permanent precondition failure from the transient server errors the retry was designed for.

It self-heals only if `chelonia/connect` is eventually called (the retry then succeeds), which is why tests (which stub `setFilter`) and Group Income (which connects at boot) never hit it.

Suggested fix — treat "no pubsub yet" as "nothing to flush" rather than a failure, since `chelonia/connect` already re-applies the complete filter set from `kvActiveFilters` (`src/chelonia.ts:1242-1246`):

```ts
// in flushDirtyFilters, before the setFilter call (or at loop entry):
if (!ctx.pubsub) {
  // chelonia/connect re-seeds client.kvFilter from kvActiveFilters,
  // so pre-connect flushes are safely dropped.
  ctx.kvFilterDirty.clear()
  return
}
```

Alternatively guard inside `chelonia/kv/setFilter` itself (`if (!this.pubsub) return`). Also consider `.unref?.()` on `kvFilterRetryTimer` so a genuinely-retrying timer cannot pin a Node process open.

## 2. 🟡 `docs/api.md` contradicts the implemented `'error'`-status seeding behavior of `chelonia/kv/update`

- [x] Addressed
- [ ] Dismissed

`docs/api.md` (added in this PR, "rejection taxonomy" section) states:

> When a slot is in `'error'` status, `update` seeds the reducer the same way `read` does: from the declared default, not from the retained mirror value. The mirror may still hold the last valid value and etag after a validation failure, but the reducer does not observe that retained value until a successful load, sync, remote update, or local write transitions the slot out of `'error'`.

This describes an earlier design. The implementation (and AGENTS.md's "Data-loss-guard reload (status quiet)" section) do the opposite for the retained-value case: `chelonia/kv/update` on an `'error'` slot that still holds a value performs a silent authoritative reload and then seeds the reducer **from the mirror** — the live server value on reload success, or the retained value on reload failure (`src/kv.ts:2492-2555`, `computeSeedValue` at `src/kv.ts:365-381`):

```ts
if (mirrorEntry && mirrorEntry.status === KV_LOAD_STATUS.ERROR && mirrorEntry.value !== undefined) {
  reloadRan = true
  ... // silent _loadSlotNow; on failure, seed from retained value
}
const seedValue = computeSeedValue(slot, mirrorEntry, reloadRan)
```

Only the `'error'`-with-no-retained-value case seeds from the default. The implemented behavior is the right one (the doc's described behavior is exactly the silent-data-loss scenario the reload guard exists to prevent), so the fix is to rewrite this `api.md` paragraph to match AGENTS.md — e.g. "When a slot is in `'error'` status but still holds a retained value, `update` first performs one silent authoritative reload and seeds the reducer from the refreshed (or retained) mirror value; only an `'error'` slot with no retained value seeds from the declared default."

While editing docs: AGENTS.md's config table says `autoLoad: 'on-demand'` "waits for `read`/`sync`" — `chelonia/kv/read` never triggers a fetch (it synchronously returns the default); only `sync` or a successful `update` materializes the value. `docs/api.md`'s selector table has the same implication for `read`. Both should say "waits for `sync` (or a successful `update`)".

## 3. ⚪️ Released contracts leave a stale empty `kvFilter` entry on the pubsub client, silencing raw-API KV frames after re-sync

- [x] Addressed
- [ ] Dismissed

`chelonia/kv/_cleanupContractRuntime` (`src/kv.ts:1910-1936`) queues an empty-filter flush before dropping per-contract state, so the flush emits `setFilter(cID, [])`. On the release path, `chelonia/pubsub/update` unsubscribes the channel synchronously (during the `CONTRACTS_MODIFIED` emit) before the flush's microtask runs, so `setKvFilter` (`src/pubsub/index.ts:977-1000`) takes the not-subscribed branch: it does not send a frame but it does record `client.kvFilter.set(cID, [])` — and an empty array means "deliver **no** KV keys". That map entry is never removed.

If the same contract is later re-synced, the SUB payload carries the stale `kvFilter: []` (`pubPayloadFactory`, `src/pubsub/index.ts:272-275`). When active slots reconcile, the subsequent flush corrects it (via the SUB-ack mismatch path), so the slot API is unaffected. But if no slot matches the re-synced contract anymore — slot definitions removed via `defineContract` replacement, or every `match` returns false — the stale `[]` persists and the server delivers no KV frames for that contract, whereas a never-slotted contract receives all of them. A raw-API consumer (the legacy `NOTIFICATION_TYPE.KV` callback) would silently observe nothing.

The "Filter ownership: don't mix slots with raw `setFilter`" section of `docs/api.md` partially covers this, but the ownership claim there is per-contract-*with-declared-slots*, while this leak outlives the slots. Suggested fix: on the release path, remove the client-side filter entry instead of setting it empty — e.g. have `_cleanupContractRuntime` skip the flush-and-empty dance when the contract is leaving the subscription set entirely, and instead call `chelonia/kv/setFilter(cID)` (the documented remove-filters form) or delete from `client.kvFilter` directly, restoring the "no filter = all keys" default for a future unslotted re-subscribe.

## 4. ⚪️ `update`/`clear` aborted after the server write commits leave the mirror stale indefinitely (by design, but undocumented for consumers)

- [x] Addressed
- [ ] Dismissed

When the caller's `signal` aborts in the window after `chelonia/kv/set` resolves, `update` records the echo CID first and then rejects with `AbortError` (`src/kv.ts:2726-2733`; same pattern in `clear` at `src/kv.ts:3016-3031`). The write has committed server-side, its pubsub echo is deliberately suppressed, and the §4.2 abort contract ("mirror unchanged, no event") is honored — with the consequence that this client's mirror (value *and* etag) stays stale until the next remote frame, local write (whose stale `ifMatch` then round-trips through a 412), or explicit `sync`. Other clients see the new value; the aborting client does not, potentially forever on a quiet slot.

The in-code rationale (`src/kv.ts:2711-2725`) is sound and I found no better alternative that keeps the abort contract, but neither AGENTS.md's abort discussion nor `docs/api.md`'s taxonomy row ("`AbortError` — mirror is unchanged; no event fires") tells the consumer that *the write may nonetheless have been persisted* and the mirror may diverge from the server afterwards. Since aborting near the commit boundary is precisely when a caller most needs to know this, add a sentence to both: "an abort that lands after the server accepted the write leaves the write committed; the mirror intentionally does not reflect it until the next load/remote/write — call `chelonia/kv/sync` after aborting if you need the mirror reconciled."

## 5. ⚪️ Dead branch in `assertJsonShape`

- [ ] Addressed
- [x] Dismissed

`src/kv.ts:317-323`:

```ts
if (value === null || value === undefined) {
  throw new ChelErrorKvValidation(...)
}
const t = typeof value
if (t === 'string' || t === 'boolean' || t === 'undefined') return
```

`t === 'undefined'` is unreachable — `value === undefined` already threw four lines above. Harmless, but the dead condition reads as if `undefined` were accepted, which is the opposite of the invariant this function enforces. Drop `|| t === 'undefined'`.

## 6. ⚪️ `types.ts` doc comment overstates the `onconflict` falsy-abort as a behavior change

- [x] Addressed
- [ ] Dismissed

The new doc block on `ChelKvOnConflictCallback` (`src/types.ts`, "NOTE: this is a behaviour change vs. pre-KV-revamp consumers, who could only return a tuple… will now silently no-op") is inaccurate: the runtime `if (!result) return false` abort path already existed on `main` (it is unchanged context in this diff, `resolveData` in `src/chelonia.ts`). Only the *type* was narrowed before; a falsy return already silently aborted the write pre-revamp rather than "historically threw downstream". The same claim is repeated in `docs/api.md`'s migration notes ("Pre-revamp consumers that accidentally returned a non-tuple falsy value would have crashed downstream; they now silently no-op"). Both should say the type now documents pre-existing runtime behavior — otherwise a direct caller auditing per these notes will hunt for a runtime change that never happened.

---

## Notes (no action required)

- **`src/kv.ts` overall**: the implementation is unusually defensive and internally consistent — staleness guards on every await boundary, the §11.2 index invariant enforced by a dev-mode checker, non-destructive mirror reads to avoid orphaned records, deep-clone discipline on every value crossing the mirror boundary, and a sound synchronous-window re-entrancy guard for `onUpdate`. The conflict-marker demote-don't-delete logic in `_handleRemote` correctly handles both echo orderings.
- The design doc's §11.6 asks for `_assertIndexConsistent` in every test's `afterEach`; the suite instead calls it at targeted points (5 call sites). Acceptable, but a global `afterEach` would be cheap insurance.
- `chelonia/kv/get` still treats HTTP 410 as an error while `chelonia/kv/set` treats 404/410 alike as "no data" — a pre-existing asymmetry this PR neither introduced nor fixed; a released-then-GC'd contract's slot load will surface `'error'` rather than `'non-init'`.
- The `cid` field on pubsub KV frames requires a matching server-side change to be emitted; until servers send it, every remote frame on an etag-bearing slot takes the authoritative-GET path (one extra GET per frame). This is documented in `docs/api.md` and degrades gracefully, but is worth remembering when sequencing the server deploy.

</details>

corrideat:
Ready

AGENTS.md
```diff
@@ -355,6 +360,294 @@ under the `./journal` subpath (`import { defaultDiff } from
...
+
+Consumer-visible leakage: `rootState._kv` is a separate subtree from
+`rootState.contracts`, but it is projected into external stores by
+`chelonia/externalStateSetup` (alongside `rootState.contracts`).
+Listeners on `CHELONIA_KV_UPDATED` / `CHELONIA_KV_STATUS_CHANGED` and
+the Vuex-style mirror receive the changed `_kv[contractID][key]` entry
+for slot updates, while contract removal drops the full per-contract KV
+subtree. Treat it as in-band with the rest of the bookkeeping data:
+redact accordingly in consumer code if needed.
+
```
taoeffect, line 223: This is way too long to be in the AGENTS.md file.

taoeffect: I missed this same issue with the Journal section that's just above this.

Please fix that section too so that it matches the style of the rest of the document. An AGENTS.md file should be a brief TOC for the agents, not API documentation.

corrideat: The latest commit should address this by making the additions much more succinct.

src/chelonia.ts
```diff
@@ -429,12 +432,33 @@ export default sbp('sbp/selectors/register', {
...
     // first time, they are removed from pending and added to subscriptionSet
     this.pending = []
     const rootState = sbp(this.config.stateSelector)
     rootState.secretKeys = rootState.secretKeys || Object.create(null)
+    // Initialise the KV mirror lazily — see KV-REVAMPED.md §5. Created
+    // as a plain object so reactive frameworks (Vue) can observe
+    // additions via `config.reactiveSet`.
+    if (!rootState._kv) {
+      this.config.reactiveSet(rootState, '_kv', Object.create(null))
+    }
```
taoeffect: I've seen this reference to a "KV mirror" mentioned a few times but it seems to not be documented anywhere. I asked Qwen3.8-Max-Preview to investigate this and come up with some documentation to add to docs/api.md and an updated comment here and here's what it came up with:

````
Here's what I'd add. The gap is specifically the *initialization* semantics the comment calls "lazily" — KV-REVAMPED.md §5 + §4.3 explain it, but `docs/api.md`'s `rootState._kv` bullet (lines 445–448) doesn't. I'd expand that bullet:

```markdown
- **`rootState._kv`** — the KV slot mirror. Slot updates push the changed
  `_kv[contractID][key]` entry; contract removal drops the full
  per-contract KV subtree. For a `_kv`-free view, project
  `{ ...rootState, _kv: undefined }`.

  The mirror is created lazily: `chelonia/_init` sets `rootState._kv`
  only if it is not already present (so a mirror restored from a
  persistence layer is preserved, not clobbered), as a null-prototype
  plain object via `config.reactiveSet` so reactive frameworks (e.g.
  Vue) observe additions. Entries are likewise created on demand — the
  reconcile pass adds `_kv[contractID][key]` only when a slot's `match`
  first returns `true`, with `status: 'non-init'` and `value: undefined`.
  The declared `defaultValue` is never copied into the mirror eagerly; it
  is substituted at read time by `chelonia/kv/read`. Direct observers of
  `rootState._kv` must therefore treat `status`, not `value`, as the
  source of truth (`value: undefined` means "presenting the default").
```

And repoint the dangling comment at `src/chelonia.ts:456` so the reference resolves in-repo:

```js
// Initialise the KV mirror lazily — see docs/api.md ("External state
// sync" → `rootState._kv`). Created as a plain object so reactive
// frameworks (Vue) can observe additions via `config.reactiveSet`.
```

This keeps it to a short note (the behavior is minor) while making the code comment self-contained. Want me to apply both?
````

I would also consider splitting up the `docs/api.md` into multiple files, as the KV stuff should probably be its own file.

Comment (taoeffect) [reply src/chelonia.ts]: I asked Kimi K3 to improve that to explain exactly what the nature of the word "mirror" is, here's what it offered:

----

You're right — the intro says "local mirroring" (line 190) but never defines it. I'd add a short subsection right after the KV slots intro (after line 193, before the selector table), since the table and everything after lean on the word:

```markdown
### The local mirror

"Mirror" refers to `rootState._kv` — a local, in-state replica of the
server-side KV store covering every declared slot. The server remains
authoritative; the mirror is what consumers actually read. For each
`(contractID, key)` pair where a slot's `match` holds, the mirror keeps
the last server-confirmed value paired with its server etag (the value
CID) plus a load status (a `KvMirrorEntry`), so `chelonia/kv/read`
answers synchronously from local state and consumers never call
`chelonia/kv/get` for a declared slot.

The library keeps the replica in lockstep with the server through four
update channels, each surfaced as a distinct `reason` on
`CHELONIA_KV_UPDATED` / `onUpdate`: the initial fetch after contract
sync (`'load'`), pubsub pushes from other clients (`'remote'`), this
client's own successful writes (`'local'`), and re-fetches after a
websocket reconnect (`'reconnect'`). Every channel validates the value
against the slot's `schema` before it lands and pairs the value with its
etag, so the mirror never holds a value the server didn't confirm.
Because the mirror lives inside Chelonia's root state, it is projected
into external stores (Vuex, Pinia) by `chelonia/externalStateSetup`
with no per-key wiring — see [External state sync](#external-state-sync).

The mirror and its entries are created lazily: `chelonia/_init` creates
`rootState._kv` only if absent (so a mirror restored from persistence
is preserved, not clobbered), as a null-prototype plain object via
`config.reactiveSet` so reactive frameworks observe additions. The
reconcile pass adds a `_kv[contractID][key]` entry only when a slot's
`match` first returns `true`, with `status: 'non-init'` and
`value: undefined`; the declared `defaultValue` is never copied in
eagerly — `chelonia/kv/read` substitutes it at read time. Direct
`rootState._kv` observers must therefore treat `status`, not `value`,
as the source of truth.
```

Then slim the external-state-sync bullet (line 445) to point at it instead of duplicating the lazy-init detail: "**`rootState._kv`** — the KV slot mirror (see [The local mirror](#the-local-mirror)). Slot updates push the changed …" — rest unchanged.

corrideat: Thanks! You're right that the 'local mirror' wasn't explicitly defined and it always seemed clear to me what it was (`_kv`).

taoeffect: It seemed clear to you because you worked on this PR, lol. :P

The docs are for newcomers :)

corrideat: I realise that and appreciate your fresh look.

src/chelonia.ts
```diff
@@ -1146,6 +1170,65 @@ export default sbp('sbp/selectors/register', {
...
+          if (!msg.channelID || !msg.key) {
+            console.info('[chelonia] Discarding kv event without channelID or key')
+            return
+          }
+          if (!this.subscriptionSet.has(msg.channelID)) {
+            console.info(
+              `[chelonia] Discarding kv event for ${msg.channelID} because it's not in the current subscriptionSet`
+            )
+            return
+          }
```
taoeffect, line 1186: Should these be `.info` or `.warn` messages?

src/kv.ts
taoeffect: So should we go for it? [zod mini](https://zod.dev/packages/mini) seems compact enough and useful for this.

Seems perfect for an agent to handle. I want to close all the open comments on this PR. This seems worth addressing given how ugly and lengthy this validation code is. All these individual manual checks and exceptions, we could greatly simplify this with Zod.

src/chelonia.ts
taoeffect: It is deserving at least of a comment above it explaining the discrepancy if it's to be left as-is.

src/chelonia.ts
corrideat: If we get here, it likely means the server is sending incorrect data. OTOH, if this infrequent enough, it can be helpful to raise visibility. If this is frequent enough, it could drown other warnings or errors.

taoeffect: @corrideat can you investigate whether these are "expected normal" or not? If they're not expected/normal, these should be warnings

corrideat: They're not expected to normally happen

docs/api.md
```diff
@@ -178,11 +178,256 @@ See [`files.md`](./files.md) for the prose guide and full signatures.
...
 | Selector | Source | Purpose |
 |---|---|---|
-| `chelonia/kv/set` | `src/chelonia.ts` | `POST /kv/:contractID/:key`. Encrypts, signs, retries on `409`. |
-| `chelonia/kv/get` | `src/chelonia.ts` | `GET /kv/:contractID/:key`. Returns the decrypted value or `undefined`. |
+| `chelonia/kv/set` | `src/chelonia.ts` | `POST /kv/:contractID/:key`. Encrypts, signs, and retries conflicts on `409`/`412`. Resolves to `{ etag: string | null }`. |
+| `chelonia/kv/get` | `src/chelonia.ts` | `GET /kv/:contractID/:key`. Returns `ChelKvGetResult | null`: the parsed encrypted/unencrypted message with `.data` and `.etag`, or `null` on 404. |
 | `chelonia/kv/setFilter` | `src/chelonia.ts` | Restrict the set of KV keys subscribed to over pubsub. |
 | `chelonia/kv/queuedSet` | `src/chelonia-utils.ts` | Wrapper around `chelonia/kv/set` that serializes concurrent updates via `okTurtles.eventQueue`. Prefer this over `kv/set` for typical writes. |
 
+## KV slots
```
taoeffect, line 186: I think `KV-REVAMPED.md` (the final version), should be added to version control somewhere. Perhaps into this `docs/` folder, maybe under `docs/specs/`. Could be useful to keep other future spec docs there too.

corrideat: Good point. I've been wondering about this myself.

I noticed the LLMs referred to specific sections of KV-REVAMPED throughout, and was going to remove those references in one of the latest iterations. I ended up deciding _against_ that because it was useful context.

What I'm not sure about is whether it should be part of this repository, or if should be part of the wiki (or both?).

taoeffect: Let's keep it in the repo. We don't really have a public wiki atm.

src/chelonia.ts
```diff
@@ -1146,6 +1170,65 @@ export default sbp('sbp/selectors/register', {
...
+          const kvSlotHandler = sbp('sbp/selectors/fn', 'chelonia/kv/_handleRemote')
```
taoeffect: There are a bunch of places I've noticed where this `'sbp/selectors/fn'` style check is done for selectors under `chelonia/kv` in this file (and perhaps elsewhere). Kimi K3 says it's because `kv.ts` isn't loaded by `chelonia.ts`...

Let's get rid of them (unless there's a strong reason to keep them around). It'll simplify the code.

I asked K3 if there were any gotchas related to doing this, and here's what it said:

<details>

Yes — `import './kv.js'` in chelonia.ts makes those guards removable, and I verified the two hazards you named. The short version: the domain lock is *safe* due to import hoisting; the circular-reference risk is *currently clean but must be defended*.

## What becomes removable

- **chelonia.ts:1103, 1186-1187, 1214** — `kvSlotHandler` is statically guaranteed: kv.ts evaluates at import time, long before `chelonia/connect` wires the pubsub dispatch.
- **internals.ts:686** — also removable, transitively: internals.ts has no `./internals` subpath export and is only reachable through chelonia.ts, so kv.ts is loaded in every scenario where internals runs.
- **`rawKvHandler` guards must stay** — that's app config, unaffected.

One adjustment: the early return `if (!rawKvHandler && !kvSlotHandler) return` becomes dead code (kvSlotHandler is always truthy), so every KV frame would cost a queue invocation even with zero slots declared. Preserve the fast path with a context check instead:

```typescript
if (!rawKvHandler && !this.kvSlotsByContractID.get(msg.channelID)?.size) return
```

This matches `_handleRemote`'s no-op-on-miss semantics exactly.

## Gotcha 1: the domain lock — safe, but understand why

`chelonia.ts:3341` runs `sbp('sbp/domains/lock', ['chelonia'])` at module top-level; afterwards, `chelonia/*` registrations are **silently dropped** (warn + skip, no throw). ESM hoisting saves you: `import './kv.js'` evaluates kv.ts *completely* before chelonia.ts's body — including the lock — executes. So kv's selectors register in time regardless of where in the file you write the import. This does quietly invert index.ts's "chelonia last" convention: the slot selectors now load whenever the core loads. index.test.ts's documented stub-before-lock ordering constraints are unaffected, and kv.test.ts is entirely unaffected (it runs in its own process and never loads chelonia.ts — it stubs the core selectors, which is exactly why its `before()` hook at line 403 can check for lock accidents).

## Gotcha 2: circular references — clean today, must stay clean

kv.ts's import closure is currently leaf-only: `errors.js`, `events.js`, `kv-constants.js`, type-only `types.js`, plus externals. **None reach chelonia.ts** — so the import is acyclic today. The trap is the future, because sbp's locked-domain drop is silent. If kv.ts (or anything in its chain) ever gains a runtime import back into chelonia.ts:

1. Something imports kv.js first → kv.ts starts evaluating
2. kv.ts → chelonia.ts → chelonia's `import './kv.js'` resolves to the *partial* in-flight module (cycle, no error)
3. chelonia.ts's body runs to completion → **domain locks**
4. Control returns to kv.ts → it registers `chelonia/kv/*` → **every selector silently dropped**

Defense: keep the rule "kv.ts never imports chelonia.ts at runtime" — anything it needs from the core must flow through string selector calls (the existing pattern), never imports. An `import/no-cycle` lint rule would enforce this statically.

## The actual cost

The guards aren't free to delete — the price is the one from before: `./chelonia` subpath consumers absorb the 166KB slot layer permanently. If that tradeoff is acceptable, the change is mechanically simple: add `import './kv.js'` next to the existing side-effect imports (chelonia.ts:50-69), delete the `sbp/selectors/fn` lookups, swap the early return, and run both test processes (`npm test` runs kv.test.ts and index.test.ts separately — both matter here).

</details>

Please task either K3 or Fable or Sol with this fix/refactor.

corrideat: This is / was intentional, precisely for the reason that Kimi says. This is mainly so that you can still import a minimal `@chelonia/lib` with just the features _you_ need by using named imports (and thus avoid the rather large `kv.ts`).

taoeffect: I do not see very many situations where you wouldn't want `kv.ts` imported. It's integral to Chelonia. We want people relying mostly on the KV store (and files) for data storage, not the contract chain.

src/chelonia.ts
```diff
@@ -2582,34 +2752,55 @@ export default sbp('sbp/selectors/register', {
       encryptionKeyId,
       signingKeyId,
       maxAttempts,
-      onconflict
+      onconflict,
+      signal: callerSignal
```
taoeffect, line 2752: I asked Fable 5 to review '`chelonia/kv/set'` specifically for bugs, and it claims it found one (and some other minor issues):

<details>

Yes — I've now done the full behavioral pass (diff against main, the spec, the new tests, and the actual server implementation in `../chel`). The core loop is solid and the new behaviors are deliberately pinned by tests, but I found one real bug worth raising, one wrong comment, and a couple of smaller notes.

### The real bug: conflict resolution can blow up on a "height-ahead" conflict body

On 409/412 the server returns the **currently stored KV value** as the response body (`chel/src/serve/routes.ts:974,987`), and `resolveData` parses it via `parseEncryptedOrUnencryptedMessage` (`src/chelonia.ts:2810`). That parse **eagerly** validates the embedded height against our locally-synced contract height and throws `ChelErrorInvalidMessageHeight` if the stored value is newer (`src/chelonia.ts:3135-3139`).

Concrete scenario: server contract height is 10, our client has synced to 8. Another client wrote the KV key after those contract ops, so the stored value carries height 9 or 10. We call `chelonia/kv/update` → POST with height 8 → server 409s (it requires an exact height match, `routes.ts:986`) → `resolveData` parses the returned value → height 10 > 8 → **throws immediately**. Consequences:

- The retry loop is short-circuited on the *first* conflict — no `onconflict`, no backoff, no chance for the parallel pubsub sync to catch us up (which is exactly what the 0–1500ms jittered retries would otherwise allow).
- `kv/update`'s catch block only maps `KvNoopAbort` and `ChelErrorKvMaxAttempts`; everything else is rethrown verbatim (`src/kv.ts:2693`). So the caller gets `ChelErrorInvalidMessageHeight`, which is **not in the §4.2 rejection taxonomy** — nobody handles it, and `kv.ts` contains zero height-related handling.
- The same eager parse affects `kv/get` and the new pubsub KV handler: a frame for a value written ahead of our sync throws during parse, gets logged and dropped, and `_handleRemote` never runs — the mirror stays stale until a reconnect or resync, with no retry hook.

Note the recovery GET can't help here — it would return the same too-new bytes. A fix probably needs to treat a height-ahead conflict body as "current value unavailable" (let `onconflict` run with `currentData: undefined`) or trigger a sync before retrying. The parse was eager in main too, so this is technically pre-existing — but the PR builds the documented taxonomy and mirror layer on top of it, which is what turns it into a real correctness gap.

### A load-bearing comment that's wrong

The "footgun" comment at `src/kv.ts:2647-2655` claims that a never-loaded slot's write "carries no `if-match` precondition" and will silently **overwrite** an unread server value, recommending `chelonia/kv/sync` first. The server says otherwise: `kv/set` always sends `if-match` (`ifMatch || '""'`), and `chel serve` treats `'""'` as a real precondition meaning "key must not exist" — an existing value produces a **412** with the current value and etag attached (`routes.ts:970-979`), after which `onconflict` merges properly. So the dangerous behavior it warns about doesn't exist, and the prescribed workaround is unnecessary. It also directly contradicts the (correct) comment at `src/chelonia.ts:2824-2828`. Only `if-match: *` overwrites unconditionally, and the client never sends that.

### Minor notes

- `kv/set` returns `{ etag: null }` both when a retry succeeds without etag headers (pinned by the test at `chelonia-kv-set.test.ts:192`) and when `onconflict` returns falsy to abort the loop — a caller can't distinguish "wrote, no header" from "didn't write". Internal callers are safe (`kv.ts` aborts via `KvNoopAbort` throw instead), but it's worth a doc line for direct users.
- With the current server, 409/412 responses always carry an `ETag` header (even if `'""'`), so the one-shot recovery GET is purely defensive — fine, just context.

</details>

src/kv.ts
```diff
@@ -0,0 +1,3481 @@
+// KV slot API — see KV-REVAMPED.md for the full design.
+//
+// This module implements the declarative KV slot API layered on top of
+// the `chelonia/kv/{set,get,setFilter}` primitives. It also hooks into
+// `chelonia/reset` (via `_waitInFlight`), `chelonia/defineContract`
+// (via `_registerContractSlots` / `_cleanupContractSlots`), and the
+// pubsub reconnect / KV-notification paths in `chelonia.ts`.
+
```
taoeffect, line 8: I had Fable 5 review `kv.ts` for bugs, here's what it found:

<details>

# Review: `src/kv.ts` (PR #80)

Reviewed all 3,482 lines against the KV-REVAMPED.md spec, the lower-level primitives in `chelonia.ts`, the pubsub client, and the server in `chel`. The KV test suite passes on this branch (156/156). The file's internal logic is remarkably tight — the lane serialization, staleness guards, echo-CID eviction, seed-precedence, and reset draining all check out under adversarial reading.

## 🔴 1. The server never sends `cid` in KV pubsub frames — the self-echo machinery is inert

- [ ] Addressed
- [x] Dismissed

The whole §4.9 design (echo suppression, `fromConflict` markers, last-write-wins gating) keys on the frame's `cid`. The spec asserts "`createKvMessage` now includes the server data `cid` in the frame," but on `chel` master, `routes.ts:1008` calls `broadcastKV(contractID, key, payloadString)` and `server.ts:127` calls `createKvMessage(contractID, key, entry)` — the optional 4th `cid` parameter (`pubsub/index.ts:425`) is never passed. No open chel PR adds it (checked, including #124 "KV attribution").

Against today's server, every frame takes the no-cid path in `_handleRemote` (`kv.ts:2186`), so:

- **Your own write's echo is not suppressed.** It triggers an authoritative GET, a second `CHELONIA_KV_UPDATED` (with `reason: 'remote'`) and a **second `onUpdate` invocation** per local write — the spec explicitly promises self-echoes produce no second event/callback.
- Every remote write costs each subscribed client one extra GET round trip (the no-cid re-pair path).
- Never-loaded slots log the `carried no cid on a never-loaded slot` warning on every frame.

The client degrades *correctly* (no data corruption), but the PR's headline feature needs a companion server change to actually work. **Quoting pitfall for whoever writes that change:** the client records `setResult.etag` verbatim from the `x-cid` header, which the server emits *quoted* — `'x-cid': "\"${newCID}\""` (`routes.ts:1016`). The broadcast `cid` must be that exact quoted string. Broadcasting the bare CID would keep suppression permanently broken, and worse: `_handleRemote` stores the frame's `cid` as the mirror etag (`kv.ts:2293`), so an unquoted cid would make the next local write send an unquoted `if-match` that the server always rejects (`routes.ts:973`) — a guaranteed 412/extra round trip on every write that follows a remote frame.

## 🟡 2. `update` writes the reducer's output into the mirror un-cloned and resolves with the same reference

- [x] Addressed
- [ ] Dismissed

`kv.ts:2737` does `reactiveSet(entryAfter, 'value', nextValue)` and `kv.ts:2764` returns that same `nextValue`. `validateReducerOutput` doesn't clone (`assertJsonShape` returns its input; a schema's `parse` may too), so both the caller (via the resolved promise) and the reducer author hold live references into the mirror. Mutating either later corrupts the mirror in place, bypassing `reactiveSet` — exactly the footgun this module documents and defends against everywhere else: `read` deep-clones on the way out (`kv.ts:2805`), events/`onUpdate` go through `cloneForEmit` (`kv.ts:860`), and `_handleRemote` explicitly clones before its mirror write with a comment explaining why (`kv.ts:2262`). The onconflict retry path has the same exposure (`nextValue = validated`, `kv.ts:2633`). Fix: clone at the mirror write (`cloneDeep(nextValue)`); that also detaches the resolved value.

## 🟡 3. The `setFilter` retry machinery guards the wrong layer

- [x] Addressed
- [ ] Dismissed

`flushDirtyFilters`' catch → `scheduleFilterRetry` (`kv.ts:918-960`) is written for "a server error while the socket stayed up," but `chelonia/kv/setFilter` → `pubsub.setKvFilter` is synchronous fire-and-forget (`pubsub/index.ts:977`) — it cannot reject on a server failure, so this path is effectively unreachable. Meanwhile the *actual* failure mode is one layer down: a server `ERR` response to `KV_FILTER` is warn-and-drop with no retry (`pubsub/index.ts:693-696`). If widening a filter fails there, the newly-added slot keys silently receive no frames (no status, no event, `kvActiveFilters` believes the filter is live) until the next reconnect/re-SUB heals it via `isKvFilterFresh`. Low probability, but worth either wiring the ERR handler to retry or at least documenting that kv.ts's retry only covers synchronous dispatch failures.

## ⚪ 4. Incorrect comment about the never-loaded-slot "footgun"

- [x] Addressed
- [ ] Dismissed

`kv.ts:2646-2655` claims a `non-init` slot's write "carries no `if-match` precondition" and will silently overwrite a server value this client never read, advising a manual `sync` first. Against the reference server this is wrong: `kv/set` sends `if-match: '""'` when `ifMatch` is undefined (`chelonia.ts:2898`), and the server 412s unless the key is absent (`routes.ts:961-979`) — so the write can't clobber anything; it falls into `onconflict`, which re-seeds the reducer from server data. The actual behavior is *safer* than documented; the comment (and its workaround advice) documents non-existent semantics. Same class: the comment in `resolveSlotDefinition` (`kv.ts:107`) says the `defaultValue` factory "is invoked exactly once here" — it's actually invoked in `_defineSlotInternal` (`kv.ts:1225`).

## ⚪ 5. Nit: `registryKey` collision with `::` in names

- [x] Addressed
- [ ] Dismissed

`KV_KEY_SEPARATOR` is `'::'`, server KV keys may contain `::` (`KV_KEY_REGEX` at `routes.ts:36` allows any non-NUL chars), and `ACTION_REGEX`'s `[^/]+` permits colons in contract type names — so `registryKey('foo/a::b', 'c') === registryKey('foo/a', 'b::c')`, aliasing two distinct registrations in `kvSlots`. Echo keys are immune (contract IDs are base58 CIDs, no colons). Pathological, but cheap to close with a collision-proof join or by rejecting `::` in keys at `defineSlot`.

</details>

src/chelonia.ts
corrideat: Ok. From your previous comment

> The guards aren't free to delete — the price is the one from before: ./chelonia subpath consumers absorb the 166KB slot layer permanently. If that tradeoff is acceptable, the change is mechanically simple: add import './kv.js' next to the existing side-effect imports (chelonia.ts:50-69), delete the sbp/selectors/fn lookups, swap the early return, and run both test processes (npm test runs kv.test.ts and index.test.ts separately — both matter here).

I'll go ahead an delete the guards as per your preference.

(Also doing with with Fable as per your request, but this is a pretty simple task IMHO)

src/chelonia.ts
corrideat: The race condition looks real, although I'm not sure it'd be very likely to happen in reality.

taoeffect: In my experience, the "anything that can go wrong will go wrong" maxim quite often holds for software development.

src/kv.ts
corrideat: @taoeffect Now, this is the only remaining item. Personally, I'd do this as a follow up issue (that we could take immediately after merging this) due to the already existing complexity of this PR.

taoeffect: Fair call 👍 

I'll open and assign an issue for it

taoeffect: Done! https://github.com/okTurtles/libcheloniajs/issues/87

corrideat:
I've now addressed all of your feedback and this should be good to merge. Just in case, I'll run the test suite from GI. ETA: test suite passed

src/kv.ts
taoeffect: Noting for reference that the first issue Fable found was for Chel, which is being tracked here: https://github.com/okTurtles/chel/issues/151

taoeffect (approved):
Approved! 😄 🎉 

Amazing work @corrideat! This is another one of those massive PRs that really pushes everything forward! Congrats!

PR #92, opened 2026-08-10
Patch on redactions

### LINKED ISSUE #82: Empty journal patch for redacted values
### Problem

If a value is redacted to a fixed value (e.g., a value is replaced by `XXXX`), it shows up in the journal log as an empty patch (or, if there were other changes, then this change is missing).

Although it makes sense since the reconstructed state from the journal won't really change, it also hides useful debugging information, as it makes it hard to know if the value is just redacted or if the action failed.

### Solution

Entries that are redacted should still show up in the logs, even if the change is effectively a no-op from the journal perspective (e.g., the value would should up as a replace operation from `XXXX` to `XXXX`)

corrideat:
Closes #82

## AI disclosure

Generated using Claude Opus 5.0 and a second round of changes applied using Qwen-3.8 Max.

Subsequent commits used Qwen-3.8 Max and GPT 5.6 Sol.


corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/31347426968)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/31636859616)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/31637012656)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/32293938226)._

**The review job timed out.**

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/32538669915)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/32595312334)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/32668830615)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/32670851265)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/32675502029)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/33283574185)._

corrideat:
It's lot of changes in this PR so I feel like I need to do a second round looking through changes.

That said, it looks fine and testing it with GI also shows it working as expected.

corrideat:
LGTM

src/journal.ts
```diff
@@ -528,18 +876,208 @@ export function shortHashRedactor (value: unknown): string {
...
+  after: RedactionSite
+): boolean {
+  const visiblePaths = defaultDiff(before.replacement, after.replacement)
+  if (visiblePaths.some((p) => p.path === '')) return false
+  const visible = new Set(visiblePaths.map((p) => p.path))
+  for (const { path } of defaultDiff(before.original, after.original)) {
+    if (!visible.has(path)) return true
+  }
+  return false
+}
```
akhileshthite, line 850: When `visiblePaths` is empty, this still runs `defaultDiff` on the unredacted originals, which clones the hidden value into a patch we throw away. Can we return early instead?

Like
```js
if (visiblePaths.length === 0) return !structurallyEqual(before.original, after.original)
```

akhileshthite (changes requested):
@corrideat Ran it locally; all tests pass, and `dist/` matches a fresh build. Just one comment.

src/journal.ts
corrideat: Right, that's a good optimisation.

akhileshthite (dismissed):
LGTM
Thanks!

src/journal.ts
```diff
@@ -235,29 +282,55 @@ function shallowEqualPrimitives (a: unknown, b: unknown): boolean {
...
+      // Own properties only — never let the prototype chain make two
+      // states look alike.
+      if (!has(bObj, k)) return false
+      if (!structurallyEqual(aObj[k], bObj[k])) return false
+    }
+    return true
+  }
+  return shallowEqualPrimitives(a, b)
+}
+
```
taoeffect: @corrideat Would turtledash's deepEqualJSONType make sense to use here?

src/journal.ts
```diff
@@ -418,34 +491,289 @@ function applyOne (root: unknown, patch: JournalPatch): unknown {
...
+// `REDACTION_NON_JSON_SAFE_SENTINEL` for every unsafe leaf (and for cyclic
+// back-references). Only called after `isJSONSafeValue` rejected the value,
+// so allocation here is the exceptional path. MUST stay in agreement with
+// `isJSONSafeValue` on what counts as unsafe (see the note above it).
+//
+// One rejection class is repaired without a sentinel: an array's own
+// enumerable non-index keys are simply dropped, because the array is
+// rebuilt index-by-index — which is exactly what JSON persistence would
+// have done to it, only now it happens once, visibly, and with a warning.
+function normalizeToJSONSafe (v: unknown, seen: Set<object>): unknown {
```
taoeffect, line 458: What about `hashableRepresentation` from turtledash? Relevant here?

src/journal.ts
```diff
@@ -528,18 +876,214 @@ export function shortHashRedactor (value: unknown): string {
...
+// and after, which is precisely the lossy case where coverage cannot be
+// established. "Hidden" is the conservative direction: markers are
+// identity writes, so an extra one is noise while a suppressed one loses
+// the only record of the hidden change. (Contrast `coveredByPatch`, where
+// ancestor coverage IS sound: patch operations are real writes into the
+// reconstructed state, not observations about a projection.)
+//
+// Always uses `defaultDiff`, never `cfg.diff`: this asks a question
+// about RFC-6901 pointer paths within a single site, which is the same
+// vocabulary the markers themselves are emitted in.
```
taoeffect: Could you please delete this entire comment and write it yourself?

It's:

1. Incomprehensible
2. Too long

src/journal.ts
```diff
@@ -93,8 +133,20 @@ export function cloneValue<T> (v: T): T {
   // (`JSON.stringify({ a: undefined })` is `"{}"`) and matches the
   // documented "plain JSON state" contract; states that rely on
   // explicit-undefined keys must supply a custom `diff` / `applyPatch`.
+  // Array holes, by contrast, are normalized to `null` (see `readIndex`),
+  // so a clone is always dense.
```
taoeffect: Does `cloneValue` (above) need to exist or can turtledash's `cloneDeep` be used instead?

src/journal.ts
```diff
@@ -235,29 +282,55 @@ function shallowEqualPrimitives (a: unknown, b: unknown): boolean {
...
+      if (!structurallyEqual(aObj[k], bObj[k])) return false
+    }
+    return true
+  }
+  return shallowEqualPrimitives(a, b)
+}
+
 // ---------------------------------------------------------------------------
 // Apply
 // ---------------------------------------------------------------------------
```
taoeffect, line 201: Could a human-generated comment please be added to this `// Apply` section explaining what these functions are for and when they are used?

taoeffect (changes requested):
Excellent work @corrideat! Just some questions plus minor requests to improve the comments that Opus left with something a bit more understandable.

src/journal.ts
corrideat: `cloneValue` behaves significantly differently from `cloneDeep`, especially for values that cannot be represented in JSON.

src/journal.ts
corrideat: `hashableRepresentation` is mostly about having a sort-stable representation. `normalizeToJSONSafe` attempts to ensure that only types that can be represented as JSON are represented.

src/journal.ts
corrideat: Maybe. I'd say both are intended to do the same thing, but `deepEqualJSONType` is much simpler and doesn't handle as many edge cases.

corrideat: @taoeffect Need your input here. Options are:

1. Leave this as is
2. Replace `deepEqualJSONType` with this function
3. Move this function (and possibly the other ones) to a separate file
4. Something else?

taoeffect: @corrideat are you saying this function is better implemented than `deepEqualJSONType`?

It's kinda hard also for me to answer your question, as you haven't said whether `deepEqualJSONType` is sufficient for our purposes here as-is. It might not handle as many edge cases but is that a problem? If it is a problem then we should stick with this function.

corrideat: > @corrideat are you saying this function is better implemented than deepEqualJSONType?

Yes

> as you haven't said whether deepEqualJSONType is sufficient for our purposes here as-is.

To answer your question I replaced this function with a call to `deepEqualJSONType` and ran the test suite. The only tests that broke were specifically about this function.

```
▶ journal: structurallyEqual
  ✖ agrees with defaultDiff on what counts as a change (0.50566ms)
  ✔ compares own properties only (0.111669ms)
  ✖ treats non-plain containers as equal only by reference (0.188354ms)
✖ journal: structurallyEqual (1.051145ms)
```

With that said, whether it's a problem or not depends on the specific inputs we provide, which I haven't looked at (but could do a deeper analysis). However, `deepEqualJSONType` has a pretty bad handling here:

```
  for (const key in a) {
    if (!deepEqualJSONType(
      a[(key as unknown as keyof typeof a)],
      b[(key as unknown as keyof typeof b)])
    ) return false
  }
```

As you see, the iteration is done over keys on `a`, with no regard to the keys on `b`. This means that `deepEqualJSONType({}, {foo:1})` will return `true`, which is wrong. Therefore, I think we should stick to this function, and also possibly replace `deepEqualJSONType` with this function.

taoeffect: K, I opened an issue in turtledash for this: https://github.com/okTurtles/turtledash/issues/6

taoeffect:
Note: `turtledash@2.0.0` has been released and this means some of the code in this PR can be removed. Not just `deepEqualJSONType` but apparently other functions too according to @corrideat.

corrideat:
Functions from this PR that made it to `turtledash`

- `deepEqualJSONType` (called `structurallyEqual` here)
- `cloneValue`
- `safeDefine`
- `readJSONIndex` (called `readIndex` here)
- `isPlainObject`
- `shallowEqualPrimitives` (not exported)

taoeffect (approved):
Nice work! 👏

PR #95, opened 2026-08-15
NS lookup

### LINKED ISSUE #90: Username lookup
### Problem

This issue was discovered during the review of <https://github.com/akhileshthite/todomvc>.

As it turns out, even though `chel` supports usernames, there's no ergonomic way of querying the username to contract ID mapping.

Instead, one needs code like:

```js
async function lookupUsername (username) {
  const response = await fetch(`${API_URL}/name/${encodeURIComponent(username)}`)
  if (response.status === 404) return null
  if (!response.ok) throw new AuthError(`Username lookup failed: ${response.status}`)
  return response.text()
}
```

### Solution

Since this is core functionality and also something that's expected to come up often, it really should be a selector exposed by `@chelonia/lib`.

corrideat:
Closes #90 

AI disclosure: Code generated with Qwen3.8-Max, reviewed by GLM-5.3.


corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/31858340196)._

src/chelonia.ts
```diff
@@ -1918,6 +1918,39 @@ export default sbp('sbp/selectors/register', {
...
+        cache: 'no-store',
+        signal: this.abortController.signal
+      }
+    )
+    // 404 means the name was never registered; 410 means the mapping was
+    // deleted (e.g. the account was). Both are reported as `null`: from
+    // the caller's perspective there simply is no current mapping, and
+    // deliberately no ChelErrorResourceGone is thrown (unlike
+    // handleFetchResult) because names have no distinct "gone" state.
+    if (response.status === 404 || response.status === 410) return null
```
akhileshthite, line 1972: One case is missing here. The relay validates the name on lookup also, not only
on registration. In chel, `GET /name/:name` runs the same `nameSchema`, so an
invalid name gives 400, not 404.

So `nameToContractID('Alice')` throws instead of returning null. But for a
caller doing a username lookup this is the same situation as "not registered",
because that name can never exist. In TodoMVC I had to check chel's
`NAME_REGEX` on the client first to avoid this.

Should 400 return null too, or at least be mentioned in the docs?

akhileshthite (changes requested):
I have one comment, the rest of the PR looks solid.

src/chelonia.ts
corrideat: That's a great question.

> for a caller doing a username lookup this is the same situation as "not registered"

I agree in most cases. One situation where it's not the same is if you're doing this to check for availability before registration (e.g., to see if you might be able to register `Alice`). In other situations, they're the same.

Based on this:

> Should 400 return null too, or at least be mentioned in the docs?

I think it should, or at least do so by default.

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/32274871491)._

corrideat:
/review

---

_AI review [started](https://github.com/okTurtles/libcheloniajs/actions/runs/32284069018)._

akhileshthite (dismissed):
Thanks!
LGTM

corrideat:
LGTM

taoeffect (approved):

